PR #1, opened 2018-03-19
Initial syncing prototype

The-Compiler:
Synchronizes a single file via SMB. Still some FIXMEs here and there, but should be ready to merge and improve upon.


ThunderKey:
@The-Compiler according to the documentation of the [SMB URL](http://ubiqx.org/cifs/Appendix-D.html) the authdomain should be infront of the username. I've corrected it in ec383f48b16ccaf2fe2c96e3c5655d3abf12d973

ThunderKey:
Travis seems to have trouble with the keyrings. What about adding the `keyrings.alt` package? It's not secure, but that is fine for testing purposes

The-Compiler:
Good catch with the SMB URL!

I rewrote some of your code in 9b73961b1e688ccb80e709498ae81d2bcad6d470 to use [pytest fixtures](https://docs.pytest.org/en/latest/fixture.html), especially [monkeypatch](https://docs.pytest.org/en/latest/monkeypatch.html) (which does a lot less magic than `unittest.mock` and is preferable for simple cases like this).

> Travis seems to have trouble with the keyrings. What about adding the keyrings.alt package? It's not secure, but that is fine for testing purposes

That test is currently flawed anyways - it'll modify your system's keyring and ask for the password when running the testsuite (if it's not unlocked yet). Instead, the test should probably set a test backend for `keyring`: https://github.com/jaraco/keyring#write-your-own-keyring-backend - do you want to take a look at this or should I?

ThunderKey:
Thanks for the improvements!

That is a good idea with the test backend. It seems straight forward, I'll give it a go this weekend.

ThunderKey:
I've added in 1fde7245328271995cd86de11db4d7152b004ce9 a `InMemoryKeyring` for the tests which gets cleared before each test.

I'm not sure if the folder `/tests/helpers` is the best location for support/helper files. Is there a prefered location?

src/kitovu/sync/syncplugin.py
```diff
@@ -0,0 +1,46 @@
+"""Loading and handling of sychronization plugins."""
+
+import abc
+import pathlib
+import typing
+
+
```
bytinbit, line 7: in our requirements file we defined a relationship between AbstractSyncPlugin and a File object. I don't see this relationship reflected in the definition of this AbstractSyncPlugin here. Why?
The only thing I can see is "fileobj: typing.IO[bytes]".

src/kitovu/sync/syncing.py
```diff
@@ -0,0 +1,57 @@
...
+from kitovu import utils
+from kitovu.sync import syncplugin
+from kitovu.sync.plugin import smb
+
+
+def _find_plugin(url: str) -> syncplugin.AbstractSyncPlugin:
+    """Find an appropriate sync plugin for the given URL.
+
+    The URL's scheme (e.g. smb://) is equal to the plugin name.
+    """
```
bytinbit, line 18: afaik you make a central design decision here: what concrete plugin is chosen depends on its protocol. However, what if there are several plugins that are http-based?

src/kitovu/sync/plugin/smb.py
```diff
@@ -0,0 +1,143 @@
...
+
+        We can't get a hash from the server, so we use the last modified time
+        (mtime) and file size as a substitute. Also see:
+        https://en.wikipedia.org/wiki/St_mtime
+
+        The mtime resolution can differ between client/server or on different
+        file systems, so we truncate it to full seconds.
+        """
+        mtime = int(mtime)
+        return f'{size}-{mtime}'
```
bytinbit, line 124: Problem 1: what unit is "size"? Byte, KB, MB?
Problem 2, resulting from 1: with this, digest can be negative (very small file, such as plaintext or config file), which I guess we don't want. why don't we create a digest using a hash function that accepts two values, size and mtime and is thus always positive?

src/kitovu/sync/syncplugin.py
```diff
@@ -0,0 +1,46 @@
...
+        This should raise an appropriate exception if the connection failed.
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def disconnect(self) -> None:
+        """Close any open connection."""
+        raise NotImplementedError
+
+    @abc.abstractmethod
```
bytinbit, line 28: chicken/egg question: we define here two functions, create_local_digest, create_remote_digest, but the smb-implementation adds another function, create_digest, that is used by both functions which were implemented from this abstract class. Is this a wise decision (code repetition)? shouldn't we better define this differently, e.g. with a create_digest-function in this class as well?

bytinbit (commented):
see comments in code. comments mainly concern design decisions.

src/kitovu/sync/syncplugin.py
The-Compiler: I think having a `File` class only makes sense after having a `FileCache`, which the initial prototype doesn't. Probably makes sense to talk about it on Tuesday though.

src/kitovu/sync/syncing.py
The-Compiler: There are actually two decisions in there:

- Plugins are chosen based on an URL, and the user configures an URL in the config. I think this one is actually the fundamental decision here, and I think it makes sense to do it that way - probably stuff to discuss on Tuesday, though.
- The scheme is used to get the plugin. We define how the URLs in the config would look, so we can just define `moodle://` or `studentenportal://` URLs. Still, this seems mostly like an implementation detail to me, and would be easy to change to use the whole URL instead. Still, gotta start somewhere for a prototype :wink:

src/kitovu/sync/plugin/smb.py
The-Compiler: Note the digest is a string, so the `-` here is just a string separator, not a mathematical minus.

src/kitovu/sync/syncplugin.py
The-Compiler: Every plugin can decide for itself how a digest would look - e.g. for SMB we use the size and modification time, but with e.g. moodle it might be possible to use some kind of hash we can get from the server. This means every plugin needs to have a `create_local_digest` and a `create_remote_digest`, because the local digest looks different based on the plugin.

The `_create_digest` in `smb.py` is just a (private) utility function so that part of the code *isn't* duplicated in `smb.py`.

The-Compiler:
@ThunderKey I did some changes to your test code in 528f5400701e0afbbd6b989c98b012d58f0d9224 - quick overview:

- The in-memory keyring is now set via a fixture (with [teardown code](https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code)) as executing code at import time (which toplevel code does) is usually a bad idea.
- `InMemoryKeyring` now uses the `d[obj]` syntax for all operations - `.get()` and `.pop()` is usually only used for special cases (when you need a fallback value instead of an error when there was no element found)
- Imports in `test_smb.py` are now grouped into three groups ([as per PEP8](https://www.python.org/dev/peps/pep-0008/#imports)) again: stdlib / third-party / project.
- `AttributesMock` and `SharedFileMock` now use [attrs](http://www.attrs.org/) which will make them e.g. comparable and shows nicer output when showing them when debugging.
- I removed `_ensure_connected` as it's easier to just do `assert self.is_connected()`.

Let me know if you disagree with any of that, or if anything seems unclear! Merging this now so we can proceed with new PRs for future changes.

PR #2, opened 2018-03-20
Created a Dockerfile und run script

ThunderKey:
# Prerequisites:

* Install docker: https://docs.docker.com/install/

# Usage

<dl>
<dt>Start pyhton:</dt><dd>`./dockerfiles/run.sh`</dd>
<dt>Start a shell:</dt><dd>`./dockerfiles/run.sh sh`</dd>
<dt>Start any command within the current environment:</dt><dd>`./dockerfiles/run.sh YOUR_CMD`</dd>
<dt>Run all validation scripts:</dt><dd>`./dockerfiles/run.sh validate`</dd>
</dl>


The-Compiler:
Probably makes sense to have a `misc/` folder for things like this?

ThunderKey:
@The-Compiler good idea, I've moved them

misc/dockerfiles/run.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+set -e
+
+echo_green() { echo -e "\e[42m$@\e[0m"; }
+echo_red()   { echo -e "\e[101m$@\e[0m"; }
```
The-Compiler, line 6: Shell escapes are typically... difficult to do in a standardized way - here, green is unreadable and red cuts of chars:

image

Not sure if it's really worth doing this - I'd rather have a simpler script without colors but readable output instead :wink:

misc/dockerfiles/run.sh
```diff
@@ -0,0 +1,35 @@
...
+
+echo_exec() { echo_green "$@"; "$@"; }
+
+DIR="$(cd "$(dirname $0)"; pwd)"
+ROOT_DIR="$(dirname $(dirname $DIR))"
```
The-Compiler, line 11: I don't really understand what you're doing here - what's the `cd` and `pwd` for? Wouldn't something like `root_dir="$(readlink -f "${0%/*}/../..")"` or so work? (see [parameter expansion](http://wiki.bash-hackers.org/syntax/pe#substring_removal))

misc/dockerfiles/run.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# copy the files for the Dockerfile into the docker dir for the build
+for f in Pipfile Pipfile.lock; do
+  cp $ROOT_DIR/$f $DIR/$f
```
The-Compiler: The variable expansions [should be quoted](https://mywiki.wooledge.org/Quotes).

misc/dockerfiles/run.sh
```diff
@@ -0,0 +1,35 @@
...
+echo_exec docker build --build-arg "UID=$(id -u)" --build-arg "GID=$(id -g)" "$DIR" --tag kitovu
+
+run_docker() {
+  echo_exec docker run --rm -ti -v "$ROOT_DIR:/opt/project" kitovu "$@"
+}
+
+if [[ "$@" == "validate" ]]; then
+  run_docker pylint src || echo_red "mpylint failed"
+  run_docker flake8 src || echo_red "mflake8 failed"
+  run_docker mypy src   || echo_red "mmypy failed"
```
The-Compiler: mmypy -> mypy

The-Compiler: Also note that the mypy arguments already changed in #1, so this isn't ideal...

The-Compiler (changes requested):

misc/dockerfiles/run.sh
ThunderKey: The problem ist, that you can either supress all build output, which can be confusing if you have changes that take a while (like `pipenv sync`). It's not clear why it takes that long.

If you print all the build output you need to group the output so that it's clear what is done where.

You're right, colours are not perfect, but it's a simple way to group them. As long as I'm the only one using it I don't think it's necessary to over-engineer this script

misc/dockerfiles/run.sh
ThunderKey: I've tested different methods at work to handle getting the full path from the `$0` with edge cases like when it's in the `$PATH` or called directly with `./` etc. and this is in my opinion the best mixture between readability and functionality. Since `$()` is called in a sub-process it does not change the current directory of the script.

misc/dockerfiles/run.sh
ThunderKey: Added quotes

misc/dockerfiles/run.sh
ThunderKey: Fixed typo and I've updated the arguments.

It's surely not ideal, but since travis checks them with `tox` the worst case scenario would be that they would show different result, which need to be fixed anyway. When this happens one can update the docker script.

The-Compiler (approved):

The-Compiler:
Thanks!

PR #5, opened 2018-03-29
Don't use URLs to pass around connection info

The-Compiler:


.pylintrc
```diff
@@ -11,4 +11,7 @@ format=colorized
 
 [basic]
 const-rgx=[a-z_]*|[A-Z_]*
-docstring-min-length=3
\ No newline at end of file
+docstring-min-length=3
+
+[design]
+max-attributes = 10
```
The-Compiler, line 17: This was needed because the `_ConnectionInfo` class has more than 7 attributes. Seems fine to bump it up a bit, IMHO.

src/kitovu/config.py
```diff
@@ -1,16 +1,22 @@
...
+    Args:
+       plugin: The name of the plugin requesting a password.
+       identifier: An unique identifier (such as an URL) for the connection.
+    """
+    service = f'kitovu-{plugin}'
+    password: typing.Optional[str] = keyring.get_password(service, identifier)
+    if password is None:
+        # FIXME handle this in a nicer way
+        password = getpass.getpass()
+        keyring.set_password(service, identifier, password)
```
The-Compiler, line 21: This is a quick & dirty hack to make it ask for a password if none is set - so now you should be able to do e.g. `kitovu sync smb fbruhin` and enter your password to test it out!

src/kitovu/sync/syncplugin.py
```diff
@@ -13,8 +13,16 @@ class AbstractSyncPlugin(metaclass=abc.ABCMeta):
     """
 
     @abc.abstractmethod
-    def connect(self, url: str, options: dict) -> None:
-        """Connect to the given URL.
+    def configure(self, info: typing.Dict[str, typing.Any]) -> None:
```
The-Compiler, line 16: I decided to separate setting a configuration in a plugin from connecting - that way, we can also do some basic error handling easily when things are missing in the config.

tests/sync/test_smb.py
```diff
@@ -85,48 +67,61 @@ class TestConnect:
...
+        plugin.configure(info)
+        plugin.connect()
+
         assert plugin._connection.init_args == {
             'username': 'myusername',
             'password': 'some_password',
             'domain': 'myauthdomain',
             'my_name': 'my-local-host',
-            'remote_name': 'myserver.test',
+            'remote_name': 'example.com',
```
The-Compiler, line 90: I switched this to `example.com` because that's reserved for purposes like this in [RFC 2606](https://tools.ietf.org/html/rfc2606) - then I did actually read that RFC, and `.test` is also reserved. Oh well, whatever, doesn't matter much :wink:

src/kitovu/sync/plugin/smb.py
```diff
@@ -76,34 +43,57 @@ class SmbPlugin(syncplugin.AbstractSyncPlugin):
...
         self._connection: SMBConnection = None
-        self._login_info = _LoginInfo()
+        self._info = _ConnectionInfo()
 
-    def connect(self, url: str, options: typing.Dict[str, typing.Any]) -> None:
-        # FIXME custom exception for errors
-        info: _LoginInfo = _parse_url(url)
-        self._login_info = info
-        password = config.get_password(url)
+    def _password_identifier(self) -> str:
```
The-Compiler, line 48: This is needed because the `keyring` library needs some kind of string as an "username"/identifier for a connection. This was the URL before, now the plugin builds a string which is unique for each connection.

src/kitovu/config.py
bytinbit: so "if password is None": in the end you just enter a bogus password? Or you take what's already there?

src/kitovu/cli.py
```diff
@@ -27,10 +27,11 @@ def cli() -> None:
...
-@click.argument('url')
-def sync(url: str) -> None:
-    """Synchronize the given URL."""
+@click.argument('plugin')
+@click.argument('username')
+def sync(plugin: str, username: str) -> None:
+    """Synchronize with the given plugin and username."""
     try:
-        syncing.start(url)
+        syncing.start(plugin, username)
```
bytinbit, line 35: so as we discussed, the plugin is thus identified via plugin+username (and not the URL concept)?

bytinbit (approved):
approve, see comments.

src/kitovu/cli.py
The-Compiler: The plugin is only identified by the plugin name (`plugin` here, which is e.g. `'smb'`). But on the commandline, you also pass the username (e.g. `kitovu sync smb msieber`) so it knows with what username to log in.

At a later stage, all that will be part of a config file, but that doesn't exist yet.

src/kitovu/config.py
The-Compiler: `getpass.getpass()` asks the user for a password on the commandline ([docs](https://docs.python.org/3/library/getpass.html)) - so we ask the user, then store the password in the keyring (for the given plugin and user), and then use the password they entered.

PR #6, opened 2018-04-03
Added prototype for the configuration file

ThunderKey:


src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,105 @@
...
+
+from kitovu import utils
+
+
+SimpleDict = typing.Dict[str, typing.Any]
+
+
+@attr.s
+class PluginSettings:
+    """A class representing the settings of a signle plugin"""
```
bytinbit: typo, single

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,105 @@
...
+
+    plugin_type: str = attr.ib()
+    connection: SimpleDict = attr.ib()
+    syncs: typing.List[SimpleDict] = attr.ib(default=[])
+
+
+@attr.s
+class Settings:
+    """A class representing the settings of a all plugins"""
```
bytinbit: typo, "all plugins"

src/kitovu/cli.py
```diff
@@ -27,11 +29,10 @@ def cli() -> None:
 
 
 @cli.command()
-@click.argument('plugin')
-@click.argument('username')
-def sync(plugin: str, username: str) -> None:
-    """Synchronize with the given plugin and username."""
+@click.argument('config_file')
+def sync(config_file: str) -> None:
+    """Synchronize with the given configuraiton file."""
```
bytinbit: typo, "configuration"

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,105 @@
...
+    root_dir: pathlib.PurePath = attr.ib()
+    global_ignore: typing.List[str] = attr.ib()
+    plugins: typing.Dict[str, PluginSettings] = attr.ib()
+
+
+class YAMLSettingsFactory:
+    """A factory to create settings objects from files or strings.
+
+    It cannot be done in the Settings class itself because mypy does
+    not know the Settings type inside of the class.
```
bytinbit: cannot know?

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+
+    @classmethod
+    def from_file(cls, path: pathlib.PurePath) -> Settings:
```
bytinbit: in order to improve readability of the code, rename "cls" to something more easily readable - I can't derive its meaning or what it's supposed to stand for.

bytinbit (commented):
found some typos, rest looks ok

src/kitovu/sync/settings.py
ThunderKey: @bytinbit Thank you for the typos, I've changed them.

`cls` is basically the same as `self` just for class methods. It is common to use `cls` instead of `class`, because `class` is a reserved keyword and stuff like `current_class` or similiar names are too verbose to be used frequently.

ThunderKey:
I've removed the `YAMLSettingsFactory` and moved the methods to the `Settings` class itself

src/kitovu/cli.py
```diff
@@ -27,11 +29,10 @@ def cli() -> None:
 
 
 @cli.command()
-@click.argument('plugin')
-@click.argument('username')
-def sync(plugin: str, username: str) -> None:
-    """Synchronize with the given plugin and username."""
+@click.argument('config_file')
```
The-Compiler: You can add `type=pathlib.Path` here (see my comment below on `Path` vs. `PurePath`) and `click` will automatically take care of converting it ([docs](http://click.pocoo.org/5/api/#click.Parameter)).

src/kitovu/cli.py
```diff
@@ -27,11 +29,10 @@ def cli() -> None:
...
+def sync(config_file: str) -> None:
+    """Synchronize with the given configuration file."""
     try:
-        syncing.start(plugin, username)
+        syncing.start_all(pathlib.PurePath(config_file))
```
The-Compiler: Why `PurePath`? As per the [pathlib docs](https://docs.python.org/3/library/pathlib.html), `PurePath` is a rather "special" thing to use - so far, we used it where paths are "virtual", i.e. not actually backed by a local file:

> Pure paths are useful in some special cases; for example: [...] You want to make sure that your code only manipulates paths without actually accessing the OS. In this case, instantiating one of the pure classes may be useful since those simply don’t have any OS-accessing operations.

So, this here should be a `Path` rather than a `PurePath` because it's actually a file.

src/kitovu/sync/plugin/smb.py
```diff
@@ -95,7 +95,7 @@ def connect(self) -> None:
         # FIXME: Handle smb.smb_structs.ProtocolError (wrong password)
         success = self._connection.connect(server_ip, self._info.port)
         if not success:
-            raise OSError("Connection failed")
+            raise OSError(f'Connection failed to {server_ip}:{self._info.port}')
```
The-Compiler, line 98: Probably makes sense to specify a custom exception class for this, but probably not as part of this PR.

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,97 @@
...
+
+from kitovu import utils
+
+
+SimpleDict = typing.Dict[str, typing.Any]
+
+
+@attr.s
+class PluginSettings:
+    """A class representing the settings of a single plugin"""
```
The-Compiler: Probably unneeded docstring?

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,97 @@
...
+
+    plugin_type: str = attr.ib()
+    connection: SimpleDict = attr.ib()
+    syncs: typing.List[SimpleDict] = attr.ib(default=[])
```
The-Compiler: There's a pitfall here: The `[]` passed to `default` is now *the same instance* of a list for all instances of `PluginSettings`. See [this explanation](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) of a similar issue when dealing with default arguments. Instead, this should be `default=attr.Factory(list)` ([docs](http://www.attrs.org/en/stable/examples.html#defaults))

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@attr.s
+class Settings:
```
The-Compiler: I don't think it makes sense to use `attrs` here - IMHO, `attrs` should only be used for simple "data classes" - i.e. classes which only have state and no behaviour. In other words, classes like a `struct` in C, with only some simple data (and perhaps some kind of getter/converter/etc.) but with no code which actually "does something".

Thus, I think it's cleaner to simply set the attributes below "normally" in a `__init__` method, and drop `@attr.s`. We did something similar in 7b908b23f2b9c534e88c3923e50712f73faff610 which made things cleaner as well.

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,97 @@
...
+    """A class representing the settings of all plugins"""
+
+    root_dir: pathlib.PurePath = attr.ib()
+    global_ignore: typing.List[str] = attr.ib()
+    plugins: typing.Dict[str, PluginSettings] = attr.ib()
+
+    @classmethod
+    def from_yaml_file(cls, path: pathlib.PurePath) -> 'Settings':
+        """Load the settings from the specified yaml file"""
+
```
The-Compiler: It's untypical in Python to have a blank line below a docstring  inside methods (and you also don't do it consistently) - I'd propose to never have a blank line after a docstring.

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,97 @@
...
+        stream = open(path, 'r')
```
The-Compiler: `open()` should never be used without a `with` block, because you can't guarantee it's closed properly (here, the file isn't closed at all) (see the [relevant section](https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files) in the Python tutorial).

Also, path is a `pathlib` object, so if it's a `Path` instead of `PurePath`, you can do:

```python
with path.open('r') as stream:
    return cls.from_yaml_stream(stream)
```

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,97 @@
...
+        return cls.from_yaml_stream(stream)
+
+    @classmethod
+    def from_yaml_stream(cls, stream: typing.TextIO) -> 'Settings':
+        """Load the settings from the specified stream"""
+
+        data = yaml.load(stream)
```
The-Compiler: We should handle `OSError` and probably `UnicodeDecodeError` properly here. Probably okay to not do it as part of this PR, but please add a `FIXME` comment.

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,97 @@
...
+
+        required_keys = ['root-dir', 'syncs', 'plugins']
+        missing_keys = [i for i in required_keys if i not in data.keys()]
```
The-Compiler: The `data` object isn't necessarily a `dict` - for example, with an empty config file it's `None` and trying to call `keys()` will cause an `AttributeError`. However, as quickly mentioned in-person today, using something like [jsonschema](https://pypi.python.org/pypi/jsonschema) or [schema](https://pypi.python.org/pypi/schema/) or [voluptuous](https://pypi.python.org/pypi/voluptuous/) for complete validation of the config file structure is probably a much better approach, so I'm not going to check for other possible validation issues in this PR.

The-Compiler: Also, as an aside: Python [sets](https://docs.python.org/3/library/stdtypes.html#set) are really nice for this kind of thing - with that, you can do:

```python
>>> required_keys = {'root-dir', 'syncs', 'plugins'}
>>> found_keys = {'root-dir'}
>>> missing_keys = required_keys - found_keys
>>> missing_keys
{'syncs', 'plugins'}
```

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,97 @@
...
+        cls._ensure_empty(data)
+
+        return Settings(
+            root_dir=root_dir,
+            global_ignore=global_ignore,
+            plugins=plugins,
+        )
+
+    @classmethod
+    def _get_plugins(cls, raw_plugins: typing.List[SimpleDict],
```
The-Compiler: Not sure if I like the name - it makes it sound like it'd load plugin instances or so - maybe `_get_plugin_settings` instead?

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,97 @@
...
+                     raw_syncs: typing.List[SimpleDict],
+                     root_dir: pathlib.PurePath) -> typing.Dict[str, PluginSettings]:
+        """Create the PluginSettings for the specified plugins and syncs."""
+        plugins = {}
+
+        for raw_plugin in raw_plugins:
+            name = raw_plugin.pop('name')
+            print(name)
```
The-Compiler: Debug code which can probably be removed? (we should introduce proper [logging](https://docs.python.org/3/library/logging.html) into the codebase anyways, but this probably isn't very helpful output even then).

src/kitovu/sync/syncing.py
```diff
@@ -28,26 +29,36 @@ def _find_plugin(pluginname: str) -> syncplugin.AbstractSyncPlugin:
     return plugin
 
 
-def start(pluginname: str, username: str) -> None:
+def start_all(config_file: pathlib.PurePath) -> None:
+    """Sync all files with the given configuration file."""
+    settings = Settings.from_yaml_file(config_file)
+    for _plugin_key, plugin_settings in settings.plugins.items():
```
The-Compiler: If you don't need the key, you can iterate over [dict.values()](https://docs.python.org/3/library/stdtypes.html#dict.values) instead of `.items()`.

However, note that dictionaries are *unordered*, so the order in which you get the plugins (and thus the order in which files are synchronized) is undefined and can change with every invocation - we don't want that, and the typical solution is to add a `sorted()` around it - so `sorted(settings.plugins.items())`. At that point, you'll need the keys again for `sorted()`, so the first part is actually okay as-is :laughing:

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+SimpleDict = typing.Dict[str, typing.Any]
+
+
+@attr.s
+class PluginSettings:
+    """A class representing the settings of a single plugin"""
+
+    plugin_type: str = attr.ib()
```
The-Compiler: I find `plugin_type` a confusing name, as it's actually the *name* of a plugin (and not some kind of e.g. Python type).

However, that'd be confusing in a different way - I'll explain in a comment further down in the example config file.

src/kitovu/sync/settings.py
```diff
@@ -0,0 +1,97 @@
...
+                connection=raw_plugin,
+            )
+
+        for sync in raw_syncs:
+            name = sync.pop('name')
+            for plugin_usage in sync.pop('plugins'):
+                plugin_usage['name'] = name
+                plugin_usage['local-dir'] = pathlib.PurePath(
+                    plugin_usage.get('local-dir', root_dir / name))
+                plugin_usage['remote-dir'] = pathlib.PurePath(plugin_usage['remote-dir'])
```
The-Compiler: This one should stay a `PurePath`. All others probably not (but I didn't comment on each individually).

src/kitovu/sync/syncing.py
```diff
@@ -7,11 +7,12 @@
 
 from kitovu import utils
 from kitovu.sync import syncplugin
+from kitovu.sync.settings import Settings, PluginSettings
 from kitovu.sync.plugin import smb
 
 
 def _find_plugin(pluginname: str) -> syncplugin.AbstractSyncPlugin:
-    """Find an appropriate sync plugin with the given name."""
+    """Find an appropriate sync plugin with the given settings."""
```
The-Compiler, line 15: This seems wrong, no? This method still only cares about the name.

tests/assets/smb_example_config.yml
```diff
@@ -0,0 +1,26 @@
+root-dir: ~/Documents/HSR/semester_06
```
The-Compiler, line 1: So - I know we discussed a lot of this in [EPJ-39](https://jira.keltec.ch/jira/browse/EPJ-39), but now that I've read the code and tried to write a config by hand (without looking at the example too much), I noticed a lot of "bad" definitions here.

Sorry for the nitpicking here, @ThunderKey @bytinbit - but I think getting good names here is a very important usability-thing for kitovu (probably *the* most important before we have a sophisticated configuration GUI) - so I added comments below.

tests/assets/smb_example_config.yml
```diff
@@ -0,0 +1,26 @@
...
+
+global-ignore:
+  - Thumbs.db
+  - .DS_Store
+
+plugins:
```
The-Compiler: `plugins` is a bad name, and it shows in the code - the difference between a `plugin`, a `plugin_name`, a `plugin_type`, a `plugin_usage`, etc. etc. is way too subtle and confusing.

Here's what I'm thinking:

A *plugin* is a backend/handler for an "external service". `smb` is one. `moodle` is one. Whatever *Leonie* writes is one.

Thus, this shouldn't be named `plugin`. The writer of the config file doesn't define or write any plugins.

What about `connections:` instead? I feel like that'd make the code much readable as well, and as an user it makes sense to define a "connection".

tests/assets/smb_example_config.yml
```diff
@@ -0,0 +1,26 @@
...
+  - name: skripte-server
+    type: smb
```
The-Compiler: This is okay-ish. But still, what is the "type" of a connection? If we call this thing "connection", we might as well just call this `plugin:` as the user actually defines what plugin to use. On the other hand, you could argue the plugin-thing is an implementation detail which shouldn't be exposed to the user, then `type:` is better.

Not sure, just some food for thought.

tests/assets/smb_example_config.yml
```diff
@@ -0,0 +1,26 @@
...
+    # hostname: svm-c213.hsr.ch
+    # port: 445
+    # share: skripte
+    # domain: HSR
+    username: example_user
+  # ...
+
+syncs:
```
The-Compiler: I think we should still come up with a better name for this. Especially in the code, if you don't know about the config file, you have no idea what a `sync` or `syncs` should be. It feels like a word we invented, but doesn't make sense to anyone else.

Ideas:

- Repositories (what openhsr-connect calls it, but it sounds like related to git)
- Modules (as in HSR-Module - bad because of exactly that ambiguity)
- Profiles (I think that's what we called them quite early - with a configuration consisting of many profiles).
- Folders
  - Directories
- Places
  - Locations
- Lessons
  - Courses
  - Subjects
- Documents

tests/sync/test_settings.py
```diff
@@ -0,0 +1,36 @@
+import pathlib
+
+from kitovu.sync.settings import Settings, PluginSettings
+
+
+def test_load_a_sample_yaml_file():
+    settings = Settings.from_yaml_file(pathlib.PurePath('./tests/assets/smb_example_config.yml'))
```
The-Compiler: Could be a fixture? Maybe okay like this for a single test (though I like separating tests from setup/teardown as much as possible), definitely needed when there's going to be more.

tests/sync/test_settings.py
```diff
@@ -0,0 +1,36 @@
...
+
+    expected_root_dir = pathlib.PurePath('~/Documents/HSR/semester_06')
```
The-Compiler: The `~` is a literal `~` here. You should call [expanduser](https://docs.python.org/3/library/pathlib.html#pathlib.Path.expanduser) or use [Path.home()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.home). Also, the real code should of course do the same.

tests/sync/test_settings.py
```diff
@@ -0,0 +1,36 @@
...
+                }
+            ],
+        ),
+    }
+
+    assert settings.root_dir == expected_root_dir
+    assert settings.global_ignore == expected_global_ignore
+    assert settings.plugins == expected_plugins
+
+    assert settings == Settings(
```
The-Compiler: This seems to test (almost) the same thing as the three lines above.

tests/sync/test_syncing.py
```diff
@@ -7,25 +7,25 @@
 from kitovu.sync.plugin import smb
 
 
-def test_find_plugin_builtin():
-    plugin = syncing._find_plugin('smb')
-    assert isinstance(plugin, smb.SmbPlugin)
+class TestFindPlugin:
```
The-Compiler, line 13: The diff is kinda hard to read, but it looks like you only added this class, right? Why?

src/kitovu/sync/syncing.py
```diff
@@ -28,26 +29,36 @@ def _find_plugin(pluginname: str) -> syncplugin.AbstractSyncPlugin:
...
     """Sync files with the given plugin and username."""
-    plugin = _find_plugin(pluginname)
-    plugin.configure({'username': username})
+    plugin = _find_plugin(plugin_settings.plugin_type)
+    plugin.configure(plugin_settings.connection)
     plugin.connect()
 
-    path = pathlib.PurePath('/Informatik/Fachbereich/Engineering-Projekt/EPJ/FS2018/')
+    for sync in plugin_settings.syncs:
+        print(sync)
```
The-Compiler: Debug code which can probably be removed?

src/kitovu/sync/syncing.py
The-Compiler: The pylint issue on this line is some kind of issue with pylint - see https://github.com/PyCQA/pylint/issues/1698. Not sure what's the problem though, as in the meantime this should be supported in pylint.

Anyways, please add

```ini
[typecheck]
ignored-classes=_CountingAttr
```

to `.pylintrc` to make it go away for now.

The-Compiler (changes requested):
Please don't let all the comments overwhelm you - some smaller and some bigger things I noticed. Still, I think for many of those, it's better to get it right now rather than later.

src/kitovu/cli.py
ThunderKey: Ah that's nice, thanks :)

src/kitovu/cli.py
ThunderKey: You're right. When I wrote it I wasn't sure what the difference is.

src/kitovu/sync/plugin/smb.py
ThunderKey: I've added a comment to https://jira.keltec.ch/jira/browse/EPJ-51. I think that this would be a good moment to improve this Error.

src/kitovu/sync/settings.py
ThunderKey: If I'd remove it pylint wouldn't be happy. Is there a way to set a nodoc?

src/kitovu/sync/settings.py
ThunderKey: Thanks for the heads up!

src/kitovu/sync/settings.py
ThunderKey: Hmm, well isn't it just a struct with class methods to help create the objects? There aren't any instance methods.

Keeping `attrs` would make the tests simpler, because the == is implemented.

src/kitovu/sync/settings.py
ThunderKey: Oh yeah, I forgot about that. Thanks

src/kitovu/sync/settings.py
ThunderKey: I've handled the validation already in the branch feature-config-file-validation, this also aborts if its None or anything else.

Those sets sound nice :D

src/kitovu/sync/settings.py
ThunderKey: Good idea

src/kitovu/sync/syncing.py
ThunderKey: I was about to change from passing the plugin name to passing the plugin settings. After I've started the rewrite I deemed it unnecessary and reverted it and forgot about the comment.

But I have changed to passing the plugin settings in the branch feature-config-file-validation to make the validations simpler

tests/assets/smb_example_config.yml
ThunderKey: I think plugin works fine.

tests/assets/smb_example_config.yml
ThunderKey: I would be for "subjects". It's not the best but those are basically a group of all files for one subject. In the US and Canada they call it "course" but I think that is too ambiguous.

tests/assets/smb_example_config.yml
ThunderKey: You're right. What about this:
```yaml
root-dir: ~/Documents/HSR/semester_06

global-ignore:
  - Thumbs.db
  - .DS_Store

connections:
  - name: skripte-server
    plugin: smb
    # hostname: svm-c213.hsr.ch
    # port: 445
    # share: skripte
    # domain: HSR
    username: example_user
  # ...

subjects:
  - name: Engineering-Projekt
    sources:
      - connection: skripte-server
        ignore:
          - SubDir
          - example.txt
        remote-dir: Informatik/Fachbereich/Engineering-Projekt/EPJ
        # local-dir: Engineering-Projekt
```

@The-Compiler @bytinbit any thoughts?

tests/sync/test_settings.py
ThunderKey: Well it is the method I try to test... I agree that making fixtures out of setup/teardown of objects, but if we move the tested method out of the test itself I think it gets a bit confusing

tests/sync/test_settings.py
ThunderKey: That's a good hint, I've added that

tests/sync/test_syncing.py
ThunderKey: Yes, I've added the class.

I prefer to have tests grouped by what they do. This can be done either by one file per test group or by grouping them within a file.

I've prepared it so when I add tests for other methods in this module that they are separated.

tests/sync/test_settings.py
ThunderKey: If anything is different with just the assert of the settings the output is not clear (at least for me).
Thats why I've added the first three asserts, but kept the asserts of the settings if some properties would be added or the constructor would change values or something like that

The-Compiler (approved):

The-Compiler:
I don't know why (maybe because of the [PyPI change](https://mail.python.org/mm3/archives/list/pypi-announce@python.org/thread/JCNO7SWQTKY54MKOZQ22PV3TUJGRM3GR/)?), but the hashes changed, which caused builds to fail. Pushed an updated Pipfile.lock now, which seems to have helped.

Merging this now!

PR #7, opened 2018-04-03
Initial prototype of a moodle plugin

The-Compiler:


ThunderKey (approved):

Pipfile.lock
```diff
@@ -80,13 +94,6 @@
             "index": "pypi",
             "version": "==6.7"
         },
-        "configparser": {
```
ThunderKey: @The-Compiler hmmm, apperently `pipenv` doesn't like me. Can you recreate the lock?

src/kitovu/sync/plugin/moodle.py
```diff
@@ -79,7 +98,9 @@ def _list_files(self, course_path: pathlib.PurePath) -> typing.Iterable[str]:
                     if 'mimetype' not in elem:
                         continue
                     local_path = module_path / elem['filename']
```
ThunderKey: This now creates the files like: `/[section name]/[module name]/[filename]` where module is a sub-group of the sections.

Unfortunately the sections are not sorted as in the moodle, but alphabetically. The only possibility I see would be to prefix a order number, but that would re-download everything if the order would be changed....

ThunderKey (dismissed):

The-Compiler:
I refactored how the mtime is set for files (in b12fbf0de65596146e436a13f551a0feb759c921), as the previous approach had various shortcomings:

- Each plugin had to fiddle with a low-level `os.utime` (and get the previous `atime`) by itself
- Since the file object wasn't closed by `syncing.py` yet, you needed to add a `flush()` which was kind of a hack.

What I did instead is that each plugin returns an mtime in `recieve_file` (optional, `None` is also possible), and `syncing.py` takes care of applying it.

What do you think of that?

src/kitovu/sync/plugin/moodle.py
```diff
@@ -0,0 +1,151 @@
...
+                                                          elem['timemodified'])
+                    yield local_path
+
+    def list_path(self, path: pathlib.PurePath) -> typing.Iterable[pathlib.PurePath]:
+        """Get a list of all courses, or files in a course."""
+        if path == pathlib.PurePath('/'):
+            for course in self._list_courses():
+                yield pathlib.PurePath(course)
+        else:
+            # FIXME
```
The-Compiler, line 121: What's there to fix?

The-Compiler:
Merging this now even if it doesn't have tests yet, as I think this has some things important for today's review (like updating the local mtime correctly so digests actually match up).

I opened issues for the remaining things:

- [EPJ-80](https://jira.keltec.ch/jira/projects/EPJ/issues/EPJ-80) (FIXME)
- [EPJ-81](https://jira.keltec.ch/jira/projects/EPJ/issues/EPJ-81) (Tests)
- [EPJ-82](https://jira.keltec.ch/jira/projects/EPJ/issues/EPJ-82) (Auth)
- [EPJ-83](https://jira.keltec.ch/jira/projects/EPJ/issues/EPJ-83) (Sortierung)

PR #9, opened 2018-04-09
Download recursively all files

bytinbit:


src/kitovu/sync/syncing.py
```diff
@@ -34,20 +35,28 @@ def start(pluginname: str, username: str) -> None:
...
 
+    # hardcoded paths that will be replaced from entries in the config-file
     path = pathlib.PurePath('/Informatik/Fachbereich/Engineering-Projekt/EPJ/FS2018/')
+    outputpath = pathlib.Path(".") / "kitovu-output-files"
+
+    files: typing.Iterable[pathlib.PurePath] = list(plugin.list_path(path))
 
-    files = list(plugin.list_path(path))
     print(f'Remote files: {files}')
+    for item in files:
```
ThunderKey: I wouldn't make the result from the list_path to a `list` and remove the print of the `files`. This way it does not need to loop over all the items twice (or with the print even three times). We print each file in the loop which should suffice

ThunderKey (changes requested):
I've noted a minor improvement and after the tests and lints succeed I'd approve it

ThunderKey (approved):

PR #10, opened 2018-04-15
File / FileCache

bytinbit:
RSVP for the structure of the File-Class, it determines everything else.


src/kitovu/sync/filecache.py
```diff
@@ -0,0 +1,70 @@
...
+                file = json_to_file["path"]
+
+        except FileNotFoundError as error:
+            raise utils.UsageError(f"File not found, {error}")
+
+    def update_file_digest(self, file: File) -> None:
+        # write File's cached_digest and path
+        pass
+
+    def discover_changes_in_file(self, file: File) -> int:
```
The-Compiler: Various methods (e.g. this one) take a `File` object. However, to construct a `File` you need a lot of information only the `FileCache` knows but the caller doesn't (such as `local_digest_at_synctime`), so that's not going to work. I think it'd be best to pass a local file path there (i.e. the path where the file is actually stored on disk), but I don't have a clear picture of things yet, either. I'd say let's do that for now.

src/kitovu/sync/filecache.py
```diff
@@ -0,0 +1,70 @@
...
+        file_properties = {"path": relative_path_with_filename,
+                           "plugin": plugin,
+                           "digests": {
+                                "cached digest": local_digest_at_synctime,
+                                }
+                           }
+
+
+class FileCache:
+    FILECACHE: str = "filecache.json"
```
The-Compiler: It'd be good if the filename (and path) could be passed to the constructor (`__init__`) instead. That'll make it easier to test as you can use a `FileCache` with a temporary file easily.

src/kitovu/sync/filecache.py
```diff
@@ -0,0 +1,70 @@
...
+    REMOTE = 2
+    BOTH = 3
+    NONE = 4
+
+
+class File:
+    def __init__(self, local_digest_at_synctime: str,
+                 relative_path_with_filename: pathlib.PurePath,
+                 plugin: syncplugin.AbstractSyncPlugin) -> None:
+        file_properties = {"path": relative_path_with_filename,
```
The-Compiler: I don't understand what you're trying to do here. All this data should probably be saved in separate attributes (`self._relative_path_with_filename`, `self._local_digest_at_synctime`, etc.).

If you need some dict-structure like this for JSON, you can add a `to_dict()` method, but it doesn't make much sense to store it like that internally I think.

src/kitovu/sync/filecache.py
```diff
@@ -0,0 +1,70 @@
...
+                           "digests": {
+                                "cached digest": local_digest_at_synctime,
+                                }
+                           }
+
+
+class FileCache:
+    FILECACHE: str = "filecache.json"
+
+    def write_file_to_filecache(self, file: File):
```
The-Compiler: You seem to try reading/writing the JSON with every operation - that's more complexity than needed, and also won't work out (you can't just dump a file to JSON and append that to the file).

Instead, I imagine things to work something like this:

- Read the entire file cache from disk once (with `json.load`) with a method like `FileCache.read()` (shouldn't need any arguments)
- Interact with the `FileCache` by adding files to it, or comparing them, etc.
- Write it to disk (with `json.dump`) once everything is finished with a method like `FileCache.write()` (or `.store()`) - again, probably without any arguments.

src/kitovu/sync/filecache.py
```diff
@@ -0,0 +1,70 @@
...
+        file_to_json = json.dumps(file)
+        try:
+            with self.FILECACHE.open("a+") as f:  # a+ = if it exsists, append; if not, create a new one
+                f.write(file_to_json)
+        except OSError as error:
+            raise utils.UsageError(f"Could not write the file, {error}")
```
The-Compiler: I wouldn't add error handling for this right now - it's very unlikely to happen (the file cache should always be writable), and a proper exception makes things much easier to debug than a "Could not write the file" message out of context (what file?). We can still discuss whether this kind of thing is needed at all, but for a first iteration, don't worry about it.

The-Compiler (changes requested):

src/kitovu/sync/filecache.py
bytinbit: Initially I had this construct as you suggested, then I changed it, since I had the idea, that it is easy to parse a dict to json. I'll revert to the original idea then.

src/kitovu/sync/kitovu_filecache.py
```diff
@@ -0,0 +1,68 @@
...
+
+class File:
+    def __init__(self, local_digest_at_synctime: str,
+                 relative_path_with_filename: pathlib.PurePath,
+                 plugin: syncplugin.AbstractSyncPlugin) -> None:
+        self._cached_digest = local_digest_at_synctime
+        self._relative_path_with_filename: pathlib.PurePath
+        self._plugin = plugin
+        self._remote_digest = ""
+        self._local_digest = ""
```
The-Compiler: You'll probably still need a `to_dict()` method or so, which converts this to a dictionary like the one you had before. My main point is that it's cleaner to store the internal data like this.

src/kitovu/sync/kitovu_filecache.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class FileCache:
+    def __init__(self, filename: str, path_to_json: pathlib.Path):
```
The-Compiler: What's the difference between `filename` and `path_to_json`? Would it suffice to only have one path (which is the full path to the file)?

src/kitovu/sync/kitovu_filecache.py
```diff
@@ -0,0 +1,68 @@
...
+        self._FILECACHE = filename  # filecache.json
+        self._path = path_to_json
+
+    def write_filecache_to_jsonfile(self, file: typing.List[File]):
```
The-Compiler: This might work, but is not what I meant exactly earlier - basically, I'd say the `FileCache` manages itself (and the caller can add individual files), and then you can tell it "save yourself" and "restore yourself".

The-Compiler (changes requested):

src/kitovu/sync/kitovu_filecache.py
```diff
@@ -0,0 +1,113 @@
...
+                "digest": self._cached_digest}
+
+
+class FileCache:
+
+    def __init__(self, filename: pathlib.Path):
+        self._filename: pathlib.Path = filename
+        self._data: typing.Dict[pathlib.Path, File] = {}
+
+    def _compare_digests(self, remote_digest: str, local_digest: str, cached_digest: str) -> Filestate:
```
The-Compiler: You could make this a bit nicer to read with two helper variables:

```python
local_changed: bool = local_digest != cached_digest
remote_changed: bool = remote_digest != cached_digest
```

src/kitovu/sync/kitovu_filecache.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def load(self) -> None:
+        """This is called first when the synchronisation process is started."""
+        with self._filename.open("r") as f:
+            json_data = json.load(f)
+        for key, value in json_data.items():
+            digest: str = value["digest"]
+            plugin: str = value["plugin"]
+            self._data[pathlib.Path(key)] = File(local_digest_at_synctime=digest, plugin=plugin)
+            # FIXME should be of type AbstractSyncPlugin - conversion needed?
```
The-Compiler: You'll need to get the plugin instance from the name here, i.e. what happens in `syncing._find_plugin`. However, that'd introduce a circular dependency (sycing depends on filecache depends on syncing), so `_find_plugin` should be moved to another place (like the `kitovu.sync.syncplugin` module).

src/kitovu/sync/kitovu_filecache.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def update(self, path: pathlib.Path, plugin: syncplugin.AbstractSyncPlugin, local_digest_at_synctime):
```
The-Compiler: "update" usually means "change an entry which is already there" - wouldn't `insert` be a better name?

src/kitovu/sync/kitovu_filecache.py
```diff
@@ -0,0 +1,113 @@
...
+        file = File(local_digest_at_synctime=local_digest_at_synctime, plugin=plugin)
+        self._data.update(path, file)
```
The-Compiler: Calling `.update()` on a dict adds another dict to it:

```python
>>> d1 = {'one': '一', 'two': '二'}
>>> d2 = {'three': '三', 'four': '四'}
>>> d1.update(d2)
>>> d1
{'one': '一', 'two': '二', 'three': '三', 'four': '四'}
```

I think you simply want `self._data[path] = file` here?

src/kitovu/sync/kitovu_filecache.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def discover_changes(self, path: pathlib.Path, plugin: syncplugin.AbstractSyncPlugin) -> Filestate:
+        """checks if the file that is currently downloaded (path-argument) has changed
+        in comparison to the local filecache and local file)."""
```
The-Compiler: The typical format for docstrings is similar to a git commit (or email with subject/body) - a one-line summary, blank line, and then a longer explanation, all in imperative form:

```
"""Check if the file currently being downloaded has changed.

(Longer text here)
"""
```

Also see [PEP 257](https://www.python.org/dev/peps/pep-0257/) which is basically PEP8 but for docstrings.

src/kitovu/sync/kitovu_filecache.py
The-Compiler: `local_digest_at_synctime` has no type.

src/kitovu/sync/kitovu_filecache.py
```diff
@@ -0,0 +1,113 @@
...
+        # FIXME error handling to see if _data has been loaded already
+
+        # use path as index into already loaded _data
+        file: File = self._data[path]
+        cached_digest = file["digest"]
```
The-Compiler: type?

src/kitovu/sync/kitovu_filecache.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # get cached_digest, compare it to file that is synchronised
+        remote_digest = plugin.create_remote_digest(path)
+        local_digest = plugin.create_local_digest(path)
```
The-Compiler: types? :wink:

src/kitovu/sync/kitovu_filecache.py
```diff
@@ -0,0 +1,113 @@
...
+            digest: str = value["digest"]
+            plugin: str = value["plugin"]
+            self._data[pathlib.Path(key)] = File(local_digest_at_synctime=digest, plugin=plugin)
+            # FIXME should be of type AbstractSyncPlugin - conversion needed?
+
+    def update(self, path: pathlib.Path, plugin: syncplugin.AbstractSyncPlugin, local_digest_at_synctime):
+        file = File(local_digest_at_synctime=local_digest_at_synctime, plugin=plugin)
+        self._data.update(path, file)
+
+    def discover_changes(self, path: pathlib.Path, plugin: syncplugin.AbstractSyncPlugin) -> Filestate:
```
The-Compiler: Can't we get the plugin from the `File` object?

src/kitovu/sync/syncing.py
```diff
@@ -10,6 +10,7 @@
 from kitovu.sync import syncplugin
 from kitovu.sync.settings import Settings, ConnectionSettings
 from kitovu.sync.plugin import smb
+from kitovu.sync import kitovu_filecache
```
The-Compiler: Why is the file named `kitovu_filecache` by the way? For consistency, you should drop that prefix. Make sure to rename it in a git commit without any other changes though, that makes things easier to follow.

src/kitovu/sync/syncing.py
```diff
@@ -29,19 +30,23 @@ def _find_plugin(pluginname: str) -> syncplugin.AbstractSyncPlugin:
     plugin: syncplugin.AbstractSyncPlugin = manager.driver
     return plugin
 
-
```
The-Compiler: Unrelated whitespace change (two blank lines between top-level functions which aren't in a class)

src/kitovu/sync/syncing.py
```diff
@@ -29,19 +30,23 @@ def _find_plugin(pluginname: str) -> syncplugin.AbstractSyncPlugin:
...
+# synchronisation process: define here filename of FileCache / write with appdirs to disk
+
 
 def start(connection_settings: ConnectionSettings) -> None:
     """Sync files with the given plugin and username."""
     plugin = _find_plugin(connection_settings.plugin_name)
     plugin.configure(connection_settings.connection)
     plugin.connect()
+    filecache: kitovu_filecache.FileCache = kitovu_filecache.FileCache(pathlib.PurePath("filecache.json"))
+    # FIXME add path from settings instead of filecache.json
```
The-Compiler: You should be able to pass something like `pathlib.Path(appdirs.user_data_dir('kitovu')) / 'filecache.json'`.

(`Path` and not `PurePath` because it's a file on disk)

src/kitovu/sync/syncing.py
```diff
@@ -51,17 +56,51 @@ def start(connection_settings: ConnectionSettings) -> None:
...
-
+            remote_digest = plugin.create_remote_digest(item)
+            print(f'Remote digest: {remote_digest}')
+
+    # special filecache cases which FIXME here
+    # 1. remote deleted (triggers exception), local exists => REMOTE*
+    # 2. remote deleted (triggers exception), local exists AND changed (local_digest and cached_digest differ) => BOTH*
+
+            # if file doesn't exist: case 3 (normal case)
+            if not pathlib.Path(local_path / item.relative_to(remote_path)).exists():
```
The-Compiler: There's a lot of duplication with code handling `Filestate.REMOTE` here, which could be simplified. Couldn't the cache check whether the local file exists and return `Filestate.REMOTE` if it doesn't?

The-Compiler (changes requested):

src/kitovu/sync/kitovu_filecache.py
bytinbit: the method is used when A has become A', and we need to reflect this in the FileCache - hence ``update()`` or ``modify()``, the latter which you originally suggested, are more concise. I'll switch to ``modify()``, since you suggested this earlier on.

src/kitovu/sync/filecache.py
```diff
@@ -0,0 +1,115 @@
...
+
+class FileCache:
+
+    def __init__(self, filename: pathlib.Path):
+        self._filename: pathlib.Path = filename
+        self._data: typing.Dict[pathlib.Path, File] = {}
+
+    def _compare_digests(self, remote_digest: str, local_digest: str, cached_digest: str) -> Filestate:
+        local_changed: bool = local_digest != cached_digest
+        remote_changed: bool = remote_digest != cached_digest
```
The-Compiler: You could rewrite the `==` parts as well, using e.g. `elif remote_changed and not local_changed:` (case 4) or `if not remote_changed and not local_changed:` (case 5), etc.

src/kitovu/sync/filecache.py
```diff
@@ -0,0 +1,124 @@
...
+
+        for key, value in self._data.items():
+            json_data[str(key)] = value.to_dict()
+
+        with self._filename.open("w") as f:
+            json.dump(json_data, f)
+
+    def load(self) -> None:
+        """This is called first when the synchronisation process is started."""
+        with self._filename.open("r") as f:
```
The-Compiler: This currently crashes when the filecache doesn't exist initially.

src/kitovu/sync/filecache.py
```diff
@@ -0,0 +1,124 @@
...
+                         local_full_path: pathlib.Path,
+                         remote_full_path: pathlib.PurePath,
+                         plugin: syncplugin.AbstractSyncPlugin) -> FileState:
+        """Check if the file that is currently downloaded (path-argument) has changed.
+
+        Change is discovered between local file cache and local file."""
+        if not local_full_path.exists():
+            return FileState.NEW
+
+        file: File = self._data[local_full_path]
```
The-Compiler, line 145: We should probably handle the case where the local file exists (because the user downloaded it manually or whatever) but it's not in the cache. Not sure how to handle that, though.

The-Compiler (changes requested):

src/kitovu/sync/filecache.py
```diff
@@ -0,0 +1,124 @@
...
+        cached_digest: str = file["digest"]
```
The-Compiler: This should be `file.cached_digest` probably? (as it's a `File` object, not a dict)

The-Compiler (changes requested):

src/kitovu/sync/filecache.py
bytinbit: I've added this as issue EPJ-79 to our backlog.

src/kitovu/sync/filecache.py
bytinbit: fixed in syncing.py

src/kitovu/sync/filecache.py
bytinbit: true, fixed.

The-Compiler (approved):

The-Compiler:
Oh, looks like there's some issues with tests on Windows, and pylint/mypy. Will fix them tomorrow, but please take a look whenever you have the time to get some free code feedback from those tools :wink:

The-Compiler:
Fixed everything up and merged! :tada: Please take a look at the commits I pushed above when you get the time, but it's mostly small stuff.

PR #12, opened 2018-04-19
Feature config file validation

ThunderKey:
`jsonschema` itself would only raise the first error. With my custom validator I've created a merged error, which would display all errors.

The `kitovu validate` command prints all errors of the base document (missing root-keys or invalid formats) and aborts. If there are no errors in the base document it prints all errors for each plugin.

I've recreated the Pipfile.lock file. It removed a few packages and it still works without them. @The-Compiler are these necessary? Especially it's a bit strange that configparser was added twice


The-Compiler:
Hmm, not sure what's up with `Pipfile.lock`. The dependencies which were removed are all things which shouldn't matter to us I think - they all look like backports of Python 3 libraries to Python 2, i.e. they aren't needed in our scenario. I *think* this just happened because of a newer Pipenv version (I can reproduce it locally as well), but I can't find a changelog.

What's confusing me more is that AppVeyor is failing because the `pytest-cov` plugin doesn't get installed?! I'll try to look into that.

The-Compiler:
Whoops, I was on the wrong branch. I did just rerun `pipenv lock` with pipenv 11.10.0 (the latest release) and those dependencies come back. Going to push that change now. Maybe you're running an older version though?

The-Compiler:
Hmm, looks like that fixed the issue on AppVeyor too ¯\_(ツ)_/¯ Going to review this now.

src/kitovu/cli.py
```diff
@@ -37,3 +38,13 @@ def sync(config: typing.Optional[pathlib.Path] = None) -> None:
...
         raise click.ClickException(str(ex))
+
+
+@cli.command()
+@click.option('--config', type=pathlib.Path, help="The configuration file to validate")
+def validate(config: typing.Optional[pathlib.Path] = None) -> None:
+    """Validates the specified configuration file."""
+    error = syncing.config_error(config)
+    if error:
+        print(error)
```
The-Compiler: You should probably do `print(error, file=sys.stderr)` here so it goes to standard error instead of standard output.

src/kitovu/sync/plugin/smb.py
```diff
@@ -134,3 +134,19 @@ def list_path(self, path: pathlib.PurePath) -> typing.Iterable[pathlib.PurePath]
 
     def retrieve_file(self, path: pathlib.PurePath, fileobj: typing.IO[bytes]) -> None:
         self._connection.retrieveFile(self._info.share, str(path), fileobj)
+
+    def connection_schema(self) -> typing.Dict[str, typing.Any]:
+        return {
+            'type': 'object',
+            'properties': {
+                'hostame': {'type': 'string'},
```
The-Compiler: hostame -> hostname

src/kitovu/sync/plugin/smb.py
```diff
@@ -134,3 +134,19 @@ def list_path(self, path: pathlib.PurePath) -> typing.Iterable[pathlib.PurePath]
...
+                'port': {'type': 'string'},
```
The-Compiler: Why string and not int?

src/kitovu/sync/plugin/smb.py
The-Compiler: Also, we should probably have some test catching this kind of mistake.

src/kitovu/sync/plugin/smb.py
```diff
@@ -134,3 +134,19 @@ def list_path(self, path: pathlib.PurePath) -> typing.Iterable[pathlib.PurePath]
 
     def retrieve_file(self, path: pathlib.PurePath, fileobj: typing.IO[bytes]) -> None:
         self._connection.retrieveFile(self._info.share, str(path), fileobj)
+
+    def connection_schema(self) -> typing.Dict[str, typing.Any]:
+        return {
+            'type': 'object',
+            'properties': {
```
The-Compiler: This is missing some advanced settings which are handled in `smb.py`:

- `sign_options` (string; `never`, `when_supported` or `when_required`)
- `use_ntlm_v2` (bool)
- `is_direct_tcp` (bool)

src/kitovu/sync/settings.py
```diff
@@ -31,43 +31,78 @@ class Settings:
     global_ignore: typing.List[str] = attr.ib()
     connections: typing.Dict[str, ConnectionSettings] = attr.ib()
 
+    settings_schema: typing.Dict[str, typing.Any] = {
```
The-Compiler: Typically, constants are all upper-case, so this should be `SETTINGS_SCHEMA`.

src/kitovu/sync/settings.py
```diff
@@ -31,43 +31,78 @@ class Settings:
...
+        'type': 'object',
+        'properties': {
+            'root-dir': {'type': 'string'},
+            'subjects': {
+                'type': 'array',
```
The-Compiler, line 39: Should probably ensure the items are strings too (but you seem to validate this separately?)

src/kitovu/sync/settings.py
```diff
@@ -31,43 +31,78 @@ class Settings:
...
+                'items': {
+                    'type': 'object',
+                    'properties': {
+                        'name': {'type': 'string'},
+                        'plugin': {'type': 'string'},
+                    },
+                    'required': ['name', 'plugin'],
+                },
+            },
+            'global-ignore': {'type': 'array'},
```
The-Compiler: Should probably ensure the items are strings too.

tests/sync/test_syncing.py
```diff
@@ -7,32 +8,76 @@
...
+
+    def connection_schema(self) -> typing.Dict[str, typing.Any]:
+        return {
+            'type': 'object',
+            'properties': {
+                'some-required-prop': {'type': 'string'},
+                'some-optional-prop': {'type': 'string'},
+            },
+            'required': ['some-required-prop'],
+            'allowAdditional': False,
```
The-Compiler: What's the difference between this and `additionalProperties` which you used elsewhere?

src/kitovu/sync/settings.py
```diff
@@ -80,6 +115,46 @@ def _get_connection_settings(cls, raw_connections: typing.List[SimpleDict],
                 connection=raw_connection,
             )
 
+        subject_schema: typing.Dict[str, typing.Any] = {
```
The-Compiler: Why is this separate from the "main" schema? So you can actually check it against the defined connections?

src/kitovu/sync/syncing.py
```diff
@@ -7,26 +7,34 @@
...
 from kitovu import utils
-from kitovu.sync import syncplugin
+from kitovu.sync.syncplugin import AbstractSyncPlugin
 from kitovu.sync.settings import Settings, ConnectionSettings
 from kitovu.sync.plugin import smb
 
 
-def _find_plugin(pluginname: str) -> syncplugin.AbstractSyncPlugin:
+def _find_plugin(plugin_settings: ConnectionSettings,
+                 validator: typing.Optional[utils.SchemaValidator] = None) -> AbstractSyncPlugin:
```
The-Compiler: Not sure if it's a good idea to do the config validation here to be honest - I think it should be separate. For example, @bytinbit will need to get a plugin from a string for the FileCache (#10) as well, and there this would complicate things needlessly.

src/kitovu/sync/syncing.py
```diff
@@ -65,3 +73,19 @@ def start(connection_settings: ConnectionSettings) -> None:
             print(f'Local digest: {digest}')
 
     plugin.disconnect()
+
+
+def config_error(config_file: typing.Optional[pathlib.Path]) -> typing.Union[str, None]:
```
The-Compiler: Maybe rename this to `validate_config` or so?

The-Compiler: Also, not sure if it really should live here, or rather in `settings.py` or so instead?

src/kitovu/sync/syncing.py
```diff
@@ -65,3 +73,19 @@ def start(connection_settings: ConnectionSettings) -> None:
...
+    Returns either an error message or None if it's valid."""
+    try:
+        settings = Settings.from_yaml_file(config_file)
+        validator = utils.SchemaValidator(abort=False)
+        for _connection_key, connection_settings in sorted(settings.connections.items()):
+            _find_plugin(connection_settings, validator)
+        if validator.valid:
+            return None
+        return validator.error_message
+    except utils.UsageError as error:
```
The-Compiler: Why can an `UsageError` happen here? I think `UsageError` implies that the user did something wrong, so I'm not sure if we should be getting one here.

The-Compiler: Having less in the `try:` block might also help with readability, because now, it's hard to tell what part could error out here.

src/kitovu/sync/syncplugin.py
```diff
@@ -52,3 +49,8 @@ def list_path(self, path: pathlib.PurePath) -> typing.Iterable[pathlib.PurePath]
     def retrieve_file(self, path: pathlib.PurePath, fileobj: typing.IO[bytes]) -> None:
         """Retrieve the given remote file."""
         raise NotImplementedError
+
+    @abc.abstractmethod
+    def connection_schema(self) -> typing.Dict[str, typing.Any]:
```
The-Compiler: I like the idea of just letting a plugin return a schema! :+1:

src/kitovu/utils.py
```diff
@@ -22,51 +23,49 @@ def get_password(plugin: str, identifier: str) -> str:
...
+class SchemaValidator:
+    """A validator for creating and merging errors in schema definitions."""
+
+    def __init__(self, abort: bool = True) -> None:
+        self.errors: typing.List[str] = []
+        self.abort: bool = abort
+
+    def validate(self, data: typing.Any, schema: typing.Dict[str, typing.Any]) -> None:
+        """Validates the given data with the schema."""
+        for error in jsonschema.Draft4Validator(schema).iter_errors(data):
```
The-Compiler: Should this use [`jsonschema.validators.validator_for(schema)`](http://python-jsonschema.readthedocs.io/en/latest/creating/#jsonschema.validators.validator_for) so it automatically picks the right validator version?

src/kitovu/utils.py
```diff
@@ -22,51 +23,49 @@ def get_password(plugin: str, identifier: str) -> str:
...
+            self.errors.append(error)
```
The-Compiler: Instead of iterating and appending, you can do `self.errors = list(validator.iter_errors(data))`

src/kitovu/utils.py
```diff
@@ -22,51 +23,49 @@ def get_password(plugin: str, identifier: str) -> str:
...
+        if self.abort and not self.valid:
+            raise InvalidSettingsError(self)
+
+    @property
+    def valid(self) -> bool:
```
The-Compiler: `is_valid`?

src/kitovu/utils.py
```diff
@@ -22,51 +23,49 @@ def get_password(plugin: str, identifier: str) -> str:
     return password
 
 
-class Error(Exception):
+class SchemaValidator:
+    """A validator for creating and merging errors in schema definitions."""
+
+    def __init__(self, abort: bool = True) -> None:
+        self.errors: typing.List[str] = []
+        self.abort: bool = abort
```
The-Compiler: I don't understand why we need two ways of signalling errors - can't we always do what we do with `abort=True` and let the caller handle the exception? You already get multiple erors in the `InvalidSettingsErrors` exception (and the message and all), so why duplicate all that functionality again with the `valid` property and `abort=False`?

tests/sync/test_syncing.py
```diff
@@ -7,32 +8,76 @@
 from kitovu import utils
 from kitovu.sync import syncing
 from kitovu.sync.plugin import smb
+from kitovu.sync.settings import ConnectionSettings
+from kitovu.sync.syncplugin import AbstractSyncPlugin
 from helpers import dummyplugin
 
 
+class FakePlugin(AbstractSyncPlugin):
```
The-Compiler: Can't we use the `DummyPlugin` here instead of having another similar thing?

tests/sync/test_syncing.py
```diff
@@ -7,32 +8,76 @@
...
         instance = manager(namespace='kitovu.sync.plugin', name='test',
                            invoke_on_load=True)
         instance.driver = fake_plugin_obj
 
-        plugin = syncing._find_plugin('test')
+        settings = self._get_settings('test', connection={'some-required-prop': 'test'})
+        plugin = syncing._find_plugin(settings)
         assert plugin is fake_plugin_obj
 
+    def _get_settings(self, plugin_name, connection={}, subjects=[]):
```
The-Compiler: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments strikes again :wink:

The-Compiler: Also, this being needed makes the tests more complex (and less like a "microtest") - another good reason why validating the plugin settings should not happen in `_find_plugin` IMHO.

tests/sync/test_syncing.py
```diff
@@ -91,3 +136,151 @@ def test_complex_sync_all(self, tmpdir: py.path.local):
             pathlib.Path(f'{tmpdir}/syncs/sync-2/group4-file1.txt'),
             pathlib.Path(f'{tmpdir}/syncs/sync-2/group4-file2.txt'),
         ]
+
+
+class TestConfigError:
+    def test_valid_configuration(self, tmpdir: py.path.local):
```
The-Compiler: Please add a blank line above this (kind of surprised flake8/pylint don't show a warning for this!)

tests/sync/test_syncing.py
```diff
@@ -91,3 +136,151 @@ def test_complex_sync_all(self, tmpdir: py.path.local):
...
+        file_name = pathlib.Path(tmpdir / 'config.yml')
+        with file_name.open('w') as f:
```
The-Compiler: `py.path.local` is already something like `pathlib.Path` (but before that existed...). You should just be able to use `config_yml = tmpdir / 'config.yml` and then `config_yml.write_text(..., encoding='utf-8')` instead ([docs](http://py.readthedocs.io/en/latest/path.html#py._path.local.LocalPath.write_text))

(same thing in other places too)

tests/sync/test_syncing.py
```diff
@@ -91,3 +136,151 @@ def test_complex_sync_all(self, tmpdir: py.path.local):
...
+        assert syncing.config_error(file_name) is None
+
+    def test_configuration_without_a_file(self, tmpdir: py.path.local):
+        file_name = pathlib.Path(tmpdir / 'config.yml')
+        assert syncing.config_error(file_name) == f'Could not find the file {file_name}'
+
+    def test_configuration_with_an_empty_file(self, tmpdir: py.path.local):
+        file_name = pathlib.Path(tmpdir / 'config.yml')
+        with file_name.open('w') as f:
+            f.write("")
```
The-Compiler: Similar to above, you can do `config_yml.ensure()` ([docs](http://py.readthedocs.io/en/latest/path.html#py._path.local.LocalPath.ensure)) which just creates an empty file if it doesn't exist.

tests/sync/test_syncing.py
The-Compiler: I also would propose using `config_yml` or so instead of `file_name` because it's not really a filename.

src/kitovu/sync/syncing.py
```diff
@@ -65,3 +73,19 @@ def start(connection_settings: ConnectionSettings) -> None:
             print(f'Local digest: {digest}')
 
     plugin.disconnect()
+
+
+def config_error(config_file: typing.Optional[pathlib.Path]) -> typing.Union[str, None]:
+    """Validates the given configuration file.
+
+    Returns either an error message or None if it's valid."""
```
The-Compiler: Makes me wonder whether it should raise an exception object (like `InvalidSettingsError`) instead. That has several benefits:

- You can reraise it as `utils.UsageError` in `cli.py` causing click to automatically do the right thing (like writing it to stderr and not stdout :wink:)
- It can have some kind of `.errors` attribute which is much easier to test, so you don't have to parse the formatted error string again for testing.

What do you think?

The-Compiler (changes requested):

src/kitovu/sync/plugin/smb.py
ThunderKey: Good catch, I've also added tests for the smb plugin

src/kitovu/sync/plugin/smb.py
ThunderKey: Oh yeah, I forgot about those

src/kitovu/sync/settings.py
ThunderKey: These are validated later, when I have the names of the plugins to validate those as well

tests/sync/test_syncing.py
ThunderKey: Hmmm, that's weird... "allowAdditional" doesn't exist, but it worked as expected. Apperently the schemas aren't really validated^^

src/kitovu/sync/settings.py
ThunderKey: Exactly

src/kitovu/sync/syncing.py
ThunderKey: Removing that from the `_find_plugin` would lead to a bit of duplicated code. And additionally in the FileCache it shouldn't create new plugin instances (or reset anything). I would suggest we provide a dicitonary of available plugins to the FileCache. For example if a plugin gets removed from the settings the FileCache shouldn't try to load it.

src/kitovu/sync/syncing.py
ThunderKey: I've thought about that as well, but if I see a method `validate_config` I either don't expect a return value or a boolean.

Moving it to `settings.py` would lead to circular dependencies. Unless we would move `_find_plugin` to the `settings.py` as well, which is a bit unintuitive.

src/kitovu/sync/syncing.py
ThunderKey: For example if the config file doesn't exist `from_yaml_file` throws an `UsageError`. Another source for a `UsageError` is `_find_plugin` which could throw a `NoPluginError`.

I could move the except three lines to the top, but I think (at least for me) that this would be more confusing. Otherwise I could make two `try` blocks, but that would lead to 3 more lines, which are IMHO unnecessary

src/kitovu/sync/syncplugin.py
ThunderKey: Thanks. It's maybe not flexible enough for each solution, but it's clean. Maybe we could make a default `validate` method which validates the data with the schema, but could be overwritten.

src/kitovu/utils.py
ThunderKey: Thanks, I didn't know this method :+1:

src/kitovu/utils.py
ThunderKey: Ah yes, makes sense^^ But I need `self.errors.extend` because `validate` can be called multiple times

src/kitovu/utils.py
ThunderKey: Alright, I'm used to the ruby version `valid?`

src/kitovu/utils.py
ThunderKey: The global Settings file should raise an exception, but the the plugins shouldn't (otherwise you'd only get the errors of the first invalid plugin).

I've started with handling it in the main method, but it just gets confusing and both the `config_error` and `start` methods would need to implement it. This way the `SchemaValidator` class is responsible for merging and preparing the error messages

tests/sync/test_syncing.py
ThunderKey: WTF? Pyhton doesn't initialize the default arguments for each method call? o.O

I think a plugin shouldn't be able to load if the settings are invalid. But I guess thats a discussion point for the next meeting

tests/sync/test_syncing.py
ThunderKey: The reason was that I wanted to set the connection_schema, but I've now added an option to set it in the `DummyPlugin` constructor

tests/sync/test_syncing.py
ThunderKey: Ok, so always a new line after a class-statement?

tests/sync/test_syncing.py
ThunderKey: That makes the code cleaner, thanks.

I had to keep it in the test where the file does not exist, because it raises a different error if it's not a `Path` object

src/kitovu/sync/syncing.py
ThunderKey: Yes, that would work. In that case I guess `validate_config` would be a good method name

The-Compiler:
Phew, that merge from `master` was a pain. I *think* I got everything working.

The-Compiler (approved):

The-Compiler:
Pushed some additional changes, should be ready now. @ThunderKey please take a look at them whenever you get the time :smile:

PR #13, opened 2018-04-22
added reporters for printing warnings

ThunderKey:
The reporter is used for printing warnings, either directly (in the CLI) or maybe only at the end (in the UI)

Usage in the pluigins: `self.reporter.warn("my message")`

### DIFF
--- a/src/kitovu/cli.py
+++ b/src/kitovu/cli.py
@@ -24,6 +24,11 @@ from kitovu import utils
 from kitovu.sync import syncing
 
 
+class CliReporter(utils.AbstractReporter):
+    def warning(self, message: str) -> None:
+        print(message)
+
+
 @click.group()
 def cli() -> None:
     pass
@@ -34,6 +39,6 @@ def cli() -> None:
 def sync(config: typing.Optional[pathlib.Path] = None) -> None:
     """Synchronize with the given configuration file."""
     try:
-        syncing.start_all(config)
+        syncing.start_all(config, CliReporter())
     except utils.UsageError as ex:
         raise click.ClickException(str(ex))
--- a/src/kitovu/cli.py
+++ b/src/kitovu/cli.py
@@ -25,6 +25,8 @@ from kitovu.sync import syncing
 
 
 class CliReporter(utils.AbstractReporter):
+    """A reporter for printing to the console."""
+
     def warn(self, message: str) -> None:
         print(message)
 
--- a/src/kitovu/cli.py
+++ b/src/kitovu/cli.py
@@ -17,6 +17,7 @@ Why does this file exist, and why not put this in __main__?
 
 import pathlib
 import typing
+import sys
 
 import click
 
@@ -28,7 +29,7 @@ class CliReporter(utils.AbstractReporter):
     """A reporter for printing to the console."""
 
     def warn(self, message: str) -> None:
-        print(message)
+        print(message, file=sys.stderr)
 
 
 @click.group()
--- a/src/kitovu/cli.py
+++ b/src/kitovu/cli.py
@@ -25,7 +25,7 @@ from kitovu.sync import syncing
 
 
 class CliReporter(utils.AbstractReporter):
-    def warning(self, message: str) -> None:
+    def warn(self, message: str) -> None:
         print(message)
 
 
--- a/src/kitovu/sync/syncing.py
+++ b/src/kitovu/sync/syncing.py
@@ -30,16 +30,17 @@ def _find_plugin(pluginname: str) -> syncplugin.AbstractSyncPlugin:
     return plugin
 
 
-def start_all(config_file: typing.Optional[pathlib.Path]) -> None:
+def start_all(config_file: typing.Optional[pathlib.Path], reporter: utils.AbstractReporter) -> None:
     """Sync all files with the given configuration file."""
     settings = Settings.from_yaml_file(config_file)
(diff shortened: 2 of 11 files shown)


src/kitovu/cli.py
```diff
@@ -24,6 +24,11 @@
 from kitovu.sync import syncing
 
 
+class CliReporter(utils.AbstractReporter):
```
The-Compiler, line 28: Add a blank line after this (to answer your question from the other PR: Yeah, after `class Foo:` there should always be a blank line (or a docstring and then a blank line).

Not sure why pylint/flake8 aren't catching this, I *think* they are in qutebrowser.

src/kitovu/sync/syncing.py
```diff
@@ -30,16 +30,17 @@ def _find_plugin(pluginname: str) -> syncplugin.AbstractSyncPlugin:
...
     for _plugin_key, connection_settings in sorted(settings.connections.items()):
-        start(connection_settings)
+        start(connection_settings, reporter)
 
 
-def start(connection_settings: ConnectionSettings) -> None:
+def start(connection_settings: ConnectionSettings, reporter: utils.AbstractReporter) -> None:
     """Sync files with the given plugin and username."""
     plugin = _find_plugin(connection_settings.plugin_name)
+    plugin.reporter = reporter
```
The-Compiler: It's bad practice (probably more than in Ruby :laughing:) to monkeypatch things like this. Instead, we should probably pass the reporter to `_find_plugin`, and that can pass it as `invoke_args` to `stevedore.driver.DriverManager` so `stevedore` passes it to the plugin's constructor ([docs](https://docs.openstack.org/stevedore/latest/reference/index.html#stevedore.driver.DriverManager))

src/kitovu/sync/syncplugin.py
```diff
@@ -12,6 +14,8 @@ class AbstractSyncPlugin(metaclass=abc.ABCMeta):
     Every abstract method in this class is a plugin hook.
     """
 
+    reporter: utils.AbstractReporter
```
The-Compiler: This is a class instead of instance variable, so this should be set in `__init__` (also see my comment above)

tests/helpers/reporter.py
```diff
@@ -0,0 +1,10 @@
+from kitovu.utils import AbstractReporter
+
+
+class TestReporter(AbstractReporter):
+    def __init__(self, raise_errors=False):
```
The-Compiler: Wouldn't it make more sense for `raise_errors` to default to `True`?

tests/helpers/reporter.py
```diff
@@ -0,0 +1,10 @@
...
+        self.raise_errors = raise_errors
+
+    def warn(self, message):
+        if self.raise_errors:
+            raise Exception(message)
```
The-Compiler: Without `raise_errors`, shouldn't we at least print them?

src/kitovu/cli.py
```diff
@@ -24,6 +24,11 @@
...
+    def warn(self, message: str) -> None:
+        print(message)
```
The-Compiler: `file=sys.stderr`? (you'll need to adjust the test, too)

The-Compiler (changes requested):

tests/helpers/reporter.py
ThunderKey: Hmmm, I'd say in the tests I'd mostly want a reportert that just ignores anything

tests/helpers/reporter.py
ThunderKey: Printing is a bit pointless if pytest doesn't show anything. Maybe for later tests it would be useful to cache the messages

The-Compiler: pytest shows the output for failing tests (and you can launch it with `-s` to make them show for all tests). Those warnings are probably useful information for failing tests.

tests/helpers/reporter.py
The-Compiler: Ignoring unexpected warnings in tests (rather than making them fail tests) sounds like the wrong thing to do, IMHO :wink:

ThunderKey: Makes sense, I've updated it to `True`.

src/kitovu/sync/syncing.py
```diff
@@ -12,34 +12,36 @@
...
+        'smb': smb.SmbPlugin(reporter),
     }
     if pluginname in builtin_plugins:
         return builtin_plugins[pluginname]
 
     try:
         manager = stevedore.driver.DriverManager(namespace='kitovu.sync.plugin',
-                                                 name=pluginname, invoke_on_load=True)
+                                                 name=pluginname, invoke_on_load=True,
+                                                 invoke_args=(reporter))
```
The-Compiler: You just found another things that sucks in Python: The syntax for single-element tuples... `(reporter)` is not a tuple, it's the same thing as just `reporter` (or e.g. `((reporter))`).

A single-element tuple is `(reporter, )` (because it's actually the comma creating the tuple syntactically... except for `()` which is special cased to be an empty tuple).

Yes, it sucks :laughing:. I'll fix it up while merging, though.

The-Compiler (changes requested):

The-Compiler (approved):

src/kitovu/sync/syncing.py
ThunderKey: Oh yeah, that's unexpected... Thanks for fixing it 👍

PR #14, opened 2018-04-26
Created plugin documentation

ThunderKey:
I've found a way to create the entire documentation of the code, but I'm not too keen on the format of this.


The-Compiler:
Why do we have to document the whole codebase? That doesn't really make much sense to me... `AbstractSyncPlugin` is pretty much the only thing relevant to plugin authors, no?

ThunderKey:
I prefer to have one location where the complete documentation is. Even if it's just an overview of all methods and child objects. For me it's a nice entrypoint to see where I can find which parts of the code

The-Compiler:
I see your point, but this is the documentation intended for *users* of the software. From an users perspective, this is "useless noise" - or even worse: By documenting all our code at the place we point plugin authors to, we pledge to keep a stable API suitable to be used in plugins *everywhere*, which we just can't.

I'll merge this for now to not block things, but let's discuss this tomorrow (I added it to the meeting notes)

The-Compiler (approved):

PR #15, opened 2018-04-26
Initial GUI

The-Compiler:
recgif


ThunderKey (changes requested):
Very nice. Seems clean and transparent.

As soon as there are a few simple tests I'll approve it.

The-Compiler:
Soooo I accidentally ended up with 100% test coverage for the GUI code :laughing: Some platform-specific stuff in there, let's see what Travis/Appveyor say.

The-Compiler:
Bleh, tests are failing to start on Windows. Looking into it.

The-Compiler:
```
C:\projects\kitovu>C:\Python36-x64\python.exe -c "from PyQt5 import QtCore"
 
C:\projects\kitovu>.\.tox\py36\Scripts\python.exe -c "from PyQt5 import QtCore"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: DLL load failed: The specified module could not be found.
 
C:\projects\kitovu>set PATH=%PATH%;C:\Python36-x64
 
C:\projects\kitovu>.\.tox\py36\Scripts\python.exe -c "from PyQt5 import QtCore"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: DLL load failed: %1 is not a valid Win32 application.
```

The-Compiler:
Took a bit more effort than I thought, but I think this is finally ready now, with tests and all. I was mistaken when saying 100% coverage above, but I think it's pretty nicely covered for now.

ThunderKey (approved):

PR #18, opened 2018-05-02
only document the code interesting for plugin creators

ThunderKey:


CONTRIBUTING.rst
```diff
@@ -16,3 +16,11 @@ Code contributions
 
 Since kitovu is currently developed as an university project, we're unable to
 accept any outside code contributions. This will change around July 2018.
+
+Documenation
```
The-Compiler: Documenation -> Documen**t**ation

docs/reference/smb.rst
```diff
@@ -0,0 +1,5 @@
+Samba Plugin
```
The-Compiler: Samba -> SMB (SMB is the protocol, Samba is a Linux implementation of SMB)

The-Compiler (changes requested):

docs/reference/smb.rst
ThunderKey: Whoops, thanks

The-Compiler (approved):

The-Compiler:
Sorry for the delay and thanks!

PR #19, opened 2018-05-02
Feature keep the order of moodle folders

ThunderKey:
Open Issue: Test module sections with illegal chars in windows

### DIFF
--- a/src/kitovu/sync/plugin/moodle.py
+++ b/src/kitovu/sync/plugin/moodle.py
@@ -98,7 +98,8 @@ class MoodlePlugin(syncplugin.AbstractSyncPlugin):
                                                        courseid=str(course_id))
 
         for section in lessons:
-            section_path: pathlib.PurePath = course_path / section['name']
+            section_nr = "{0:0=2d}".format(section['section'])
+            section_path: pathlib.PurePath = course_path / f"{section_nr} - {section['name']}"
             for module in section['modules']:
                 module_path: pathlib.PurePath = section_path / module['name']
                 for elem in module.get('contents', []):
--- a/src/kitovu/sync/plugin/moodle.py
+++ b/src/kitovu/sync/plugin/moodle.py
@@ -98,7 +98,7 @@ class MoodlePlugin(syncplugin.AbstractSyncPlugin):
                                                        courseid=str(course_id))
 
         for section in lessons:
-            section_nr = "{0:0=2d}".format(section['section'])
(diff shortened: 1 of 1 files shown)


The-Compiler:
The branch should probably be based on `master` and not your docs branch?

The-Compiler:
> Open Issue: Test module sections with illegal chars in windows

Seems like a general issue that might appear in other plugins as well. We should probably strip some special chars from local filenames - but that should be a separate story, IMHO.

ThunderKey:
Damn it, I was in the wrong branch....

> Seems like a general issue that might appear in other plugins as well. We should probably strip some special chars from local filenames - but that should be a separate story, IMHO.

I think that's very important behaviour, otherwise it could break in the usability tests. I could add it in this PR. Do you know of a built in way? Otherwise I'll research illegal chars for linux, mac and windows and replace those.

The-Compiler:
I'd prefer having a separate PR, IMHO our PRs were a bit too big lately :wink:

Not aware of a built-in way for this, but I wrote this for qutebrowser some while ago and it seems to work well: https://github.com/qutebrowser/qutebrowser/blob/v1.2.1/qutebrowser/utils/utils.py#L490-L508

The-Compiler:
Taking the freedom to rebase and force-push this to clean things up.

src/kitovu/sync/plugin/moodle.py
```diff
@@ -98,7 +98,8 @@ def _list_files_in_course(self,
                                                        courseid=str(course_id))
 
         for section in lessons:
-            section_path: pathlib.PurePath = course_path / section['name']
+            section_nr = "{0:0=2d}".format(section['section'])
```
The-Compiler: I don't really understand the format string here - you can drop the initial `0` and just do `{:0=2d}` (no need to number fields in Python 2.7+). However, what's the `=` for? Looking at [pyformat.info](https://pyformat.info/#number_sign) it seems to control where the sign is going, but we probably don't expect negative numbers here? So I think this can just be `{:02d}` instead, right?

The-Compiler (changes requested):

src/kitovu/sync/plugin/moodle.py
ThunderKey: Thanks for the clarification. TBH the documentation of the `format` method was too confusing for me to completely understand whilst writing this code. So I used this which seemed to work ;)

The-Compiler (approved):

PR #20, opened 2018-05-03
added a sanitize_filename method

ThunderKey:


The-Compiler (changes requested):
This will break things, because `syncing.py` tries then to ask for the filenames the plugin returned, and those don't actually exist on the server.

Instead, `syncing.py` should probably only sanitize the output filename, no?

ThunderKey:
Oh yeah...

The problem would be Folders with slashes in them. This would lead to sub folders, which could be confusing. E.g. "17./18. of March"

ThunderKey:
But I guess it's the best option. The only alternative I'd see would be a "translation" dictionary, which would make everything much more complicated

src/kitovu/sync/syncing.py
```diff
@@ -77,7 +77,8 @@ def start(connection_settings: ConnectionSettings, reporter: utils.AbstractRepor
             # remote_full_path: /Informatik/Fachbereich/EPJ/Dokumente/Anleitung.pdf
             #   with relative_to: Dokumente/Anleitung.pdf
             # -> local_full_path: /home/leonie/HSR/EPJ/Dokumente/Anleitung.pdf
-            local_full_path: pathlib.Path = local_dir / remote_full_path.relative_to(remote_dir)
+            local_full_path: pathlib.Path = local_dir / \
+                utils.sanitize_filename(str(remote_full_path.relative_to(remote_dir)))
```
The-Compiler: I'd prefer to add a second variable to break this, e.g.

```python
filename: str = utils.sanitize_filename(str(remote_full_path.relative_to(remote_dir)))
local_full_path: pathlib.Path = local_dir / filename
```

src/kitovu/utils.py
```diff
@@ -24,6 +24,26 @@ def get_password(plugin: str, identifier: str) -> str:
     return password
 
 
+def sanitize_filename(name: str, replacement: typing.Optional[str] = '_') -> str:
```
The-Compiler: Might be more in line with what we do elsewhere to take a `pathlib.PurePath` (and return one) here instead.

src/kitovu/utils.py
```diff
@@ -24,6 +24,26 @@ def get_password(plugin: str, identifier: str) -> str:
...
+    r"""Replace invalid filename characters.
+
+    Note: This does not escape directory separators (/ and \).
+
+    Args:
+        name: The filename.
+        replacement: The replacement character (or None).
+    """
```
The-Compiler: Probably okay to drop at least the `Args:` part of the docstring here.

src/kitovu/utils.py
```diff
@@ -24,6 +24,26 @@ def get_password(plugin: str, identifier: str) -> str:
...
+    if replacement is None:
+        replacement = ''
```
The-Compiler: We don't need this for kitovu, so let's remove it. Actually, we don't need the ability to customize the replacement char at all.

tests/test_utils.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+
+from kitovu import utils
+
+
+class TestSanitizeFilename():
```
The-Compiler: Drop the `()` for consistency with other classes.

tests/test_utils.py
```diff
@@ -0,0 +1,29 @@
+import pytest
```
The-Compiler, line 1: Also see https://github.com/qutebrowser/qutebrowser/blob/v1.3.0/tests/unit/utils/test_utils.py#L632-L638 for my tests, but I think yours are fine as well.

The-Compiler (changes requested):

The-Compiler (approved):

The-Compiler:
I opened [EPJ-94](https://jira.keltec.ch/jira/browse/EPJ-94) for how to handle slashes in filenames. Like you say, it'd be nice to handle this in a nicer way.

PR #21, opened 2018-05-03
Feature add docs and edit cli commands

ThunderKey:
Issues:
* [x] Make `kitovu docs` version aware.
 * How could we make different versions in readthedocs?
* [x] Get the default filename in `kitovu edit`
 * Currently it is built in the `settings.py`
 * Option 1: import `settings.py` in the `cli.py` => circular dependencies
 * Option 2: Move the default filename method to `utils.py` and still set it in `settings.py` for most commands
 * Option 3: Move the default filename method to `utils.py` and always pass the filename to the `syncing.py`. Advantage: less optional args; Disadvantage: more logic in the cli

@The-Compiler, @bytinbit  what are your opinions?

### DIFF
--- a/src/kitovu/cli.py
+++ b/src/kitovu/cli.py
@@ -86,8 +86,8 @@ DEFAULT_EDITORS = [
 
 @cli.command()
 @click.option('--config', type=pathlib.Path, help="The configuration file to edit")
-@click.option('--editor', type=str, help=f"The command of the editor to use. \
-              Default: $EDITOR or the first existing out of {', '.join(DEFAULT_EDITORS)}")
+@click.option('--editor', type=str, help="The command of the editor to use. "
+              f"Default: $EDITOR or the first existing out of {', '.join(DEFAULT_EDITORS)}")
 def edit(config: typing.Optional[pathlib.Path] = None, editor: typing.Optional[str] = None) -> None:
     """Edit the specified configuration file."""
     if editor is None and 'EDITOR' in os.environ:
--- a/src/kitovu/cli.py
+++ b/src/kitovu/cli.py
@@ -18,7 +18,6 @@ Why does this file exist, and why not put this in __main__?
 import pathlib
 import typing
 import sys
-import platform
 from distutils import spawn
 import subprocess
 import os
@@ -76,7 +75,7 @@ def docs() -> None:
     webbrowser.open_new_tab('https://kitovu.readthedocs.io/en/latest')
 
 
-AVAILABLE_EDITORS = [
+DEFAULT_EDITORS = [
     'vim',
     'emacs',
     'nano',
@@ -87,7 +86,8 @@ AVAILABLE_EDITORS = [
 
 @cli.command()
 @click.option('--config', type=pathlib.Path, help="The configuration file to edit")
-@click.option('--editor', type=str, help=f"The command of the editor to use. Default: $EDITOR or the first existing out of {', '.join(AVAILABLE_EDITORS)}")
+@click.option('--editor', type=str, help=f"The command of the editor to use. \
+              Default: $EDITOR or the first existing out of {', '.join(DEFAULT_EDITORS)}")
 def edit(config: typing.Optional[pathlib.Path] = None, editor: typing.Optional[str] = None) -> None:
     """Edit the specified configuration file."""
     if editor is None and 'EDITOR' in os.environ:
@@ -108,8 +108,8 @@ def _get_editor_path(editor: typing.Optional[str]) -> typing.Optional[str]:
     if editor is not None:
         return spawn.find_executable(editor)
 
-    for e in AVAILABLE_EDITORS:
-        path = spawn.find_executable(e)
+    for default_editor in DEFAULT_EDITORS:
+        path = spawn.find_executable(default_editor)
         if path is not None:
             return path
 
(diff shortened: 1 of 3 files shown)


The-Compiler:
> * [ ] Make `kitovu docs` version aware.
> * How could we make different versions in readthedocs?

I think RTD does that automatically [once tags are pushed](http://docs.readthedocs.io/en/latest/versions.html) - see e.g. the "v: latest" thing at the bottom right on https://docs.pytest.org/.

However, I think until we actually have a first release, it's okay to link to the `/latest` URL like you did.

> * [ ] Get the default filename in `kitovu edit
> * Currently it is built in the `settings.py`
> * Option 1: import `settings.py` in the `cli.py` => circular dependencies

I don't follow - why circular? `cli` (at the top) imports `settings`, but not the other way around.

src/kitovu/cli.py
```diff
@@ -62,3 +66,51 @@ def validate(config: typing.Optional[pathlib.Path] = None) -> None:
...
+    'emacs',
+    'nano',
+    'editor',
+    'notepad',
+]
+
+
+@cli.command()
+@click.option('--config', type=pathlib.Path, help="The configuration file to edit")
+@click.option('--editor', type=str, help=f"The command of the editor to use. \
```
The-Compiler: I'd prefer ending the string there (with `"`) instead of the backslash, and then opening a new string (via `f"`) on the next line. Python will automatically concatenate them, and it's less brittle than the backslash (when e.g. adding a space or whatever).

src/kitovu/cli.py
```diff
@@ -62,3 +66,51 @@ def validate(config: typing.Optional[pathlib.Path] = None) -> None:
...
+    if editor is None and 'EDITOR' in os.environ:
+        editor = os.environ['EDITOR']
+    editor_path: typing.Optional[str] = _get_editor_path(editor)
+
+    if editor_path is None:
+        if editor is None:
+            print('Could not find a valid editor')
+        else:
+            print(f"Could not find the editor {editor}")
+        sys.exit(1)
```
The-Compiler: Raise a [ClickException](http://click.pocoo.org/5/api/#click.ClickException) instead? Maybe even from `_get_editor_path` (instead of returning `None`) instead of here.

tests/test_cli.py
```diff
@@ -15,3 +23,102 @@ def test_reporter(capsys, reporter):
     captured = capsys.readouterr()
     assert captured.out == ""
     assert captured.err == "my test\nanother test\n"
+
+
+def test_docs(runner, monkeypatch):
+    calls = []
+    monkeypatch.setattr('webbrowser.open', lambda *args: calls.append(args))
```
The-Compiler: For cases like this where you actually need to record calls, I'd use `pytest-mock` (I use `monkeypatch` for simple cases where I don't need to record things or mock complex stuff, and `mock` when I need to).

There, you can simply use `mocker.patch('webbrowser.open', autospec=True)` (IIRC) and then use `mocker.assert_called_once_with(...)`.

The-Compiler: Also, why not patch the `webbrowser` function we're actually calling? Right now the patching is relying on implementation internal stuff in `webbrowser.py`.

tests/test_cli.py
```diff
@@ -15,3 +23,102 @@ def test_reporter(capsys, reporter):
...
+    assert calls == [('https://kitovu.readthedocs.io/en/latest', 2)]
+
+
+class TestEdit:
+
+    @pytest.fixture(autouse=True)
+    def clear_env(self, monkeypatch):
+        monkeypatch.delenv('EDITOR', raising=False)
+
+    class TestMissing:
```
The-Compiler, line 42: Does nesting classes actually work fine with pytest? I've never tried it, but it's something you'd usually not do, so I'd prefer two classes named `TestEditMissing` and `TestEditAvailable` or so.

The-Compiler (changes requested):

src/kitovu/cli.py
```diff
@@ -62,3 +66,51 @@ def validate(config: typing.Optional[pathlib.Path] = None) -> None:
...
+
+    subprocess.call([editor_path, config])
```
The-Compiler: mypy complains here because it thinks `editor_path` can still be `None`. You should be able to fix it by adding a `assert editor_path is not None` above (I hope).

tests/test_cli.py
ThunderKey: Thanks for the hint with the mocker.

The idea of checking against `webbrowser.open` was to ensure that no other method of `webbrowser` would get called. But I can remove it

tests/test_cli.py
ThunderKey: It does work. The advantag is that the clear_env is called in the subclasses as well.

I personally prefer it to be grouped as much as possible, especially with fixtures and before/after hooks it's in my opinion more transparent.

src/kitovu/cli.py
ThunderKey: Adding the default value fixed it. I've no idea why I thought this would lead to a circular dependency. Maybe because I think the cli uses too many namespaces and if we would want to extract the `get_config_file_path` to prevent it, it would lead to circular dependencies.

The-Compiler (approved):

PR #23, opened 2018-05-08
Fix issues with setup.py and Pipfile

The-Compiler:
This makes sure dependencies are actually being installed when installing kitovu via
"python3 setup.py install" (or "pip install -e .").

Pipenv/Pipfile are the future™, but for now we need to support the past, too.

Also, this moves development requirements to `[dev-packages]`.


bytinbit (approved):
looks good :)

PR #24, opened 2018-05-08
Only import GUI if necessary

The-Compiler:
Otherwise, PyQt5 is required to run the CLI.


bytinbit (approved):
Approved, keeps cli-code clean.

PR #26, opened 2018-05-08
Remove reporters

The-Compiler:
Part of [EPJ-51](https://jira.keltec.ch/jira/browse/EPJ-51). ~~Based on #25 so you might want to merge that first.~~

PR #25, opened 2018-05-08
Add basic logging support

The-Compiler:
See [EPJ-74](https://jira.keltec.ch/jira/browse/EPJ-74):

```
$ kitovu sync                                                           
Syncing connection moodle
Syncing subject VSS

Syncing connection skripte-server
Syncing subject Engineering-Projekt
Downloading Informatik/Fachbereich/Engineering-Projekt/EPJ/FS2018/Projektantraege/Projektantrag_kitovu.pdf
```

```
$ kitovu --loglevel debug sync
[...]
2018-05-08 07:57:16,823 [DEBUG]       kitovu.sync.syncing Checking: Informatik/Fachbereich/Engineering-Projekt/EPJ/FS2018/Projektantraege/projektantrag_hiddencomment.pdf
2018-05-08 07:57:16,827 [DEBUG]       kitovu.sync.syncing Remote digest: 597226-1519720070
2018-05-08 07:57:16,827 [DEBUG]     kitovu.sync.filecache Discovering changes for local: /home/florian/proj/kitovu/synctest/Engineering-Projekt/FS2018/Projektantraege/projektantrag_hiddencomment.pdf / remote: Informatik/Fachbereich/Engineering-Projekt/EPJ/FS2018/Projektantraege/projektantrag_hiddencomment.pdf by plugin smb
2018-05-08 07:57:16,831 [DEBUG]     kitovu.sync.filecache Comparing digests: remote 597226-1519720070, local 597226-1519720070, cached 597226-1519720070
2018-05-08 07:57:16,832 [DEBUG]       kitovu.sync.syncing No remote changes.
2018-05-08 07:57:16,832 [DEBUG]       kitovu.sync.syncing Checking: Informatik/Fachbereich/Engineering-Projekt/EPJ/FS2018/Projektantraege/Projektantrag_kitovu.pdf
2018-05-08 07:57:16,836 [DEBUG]       kitovu.sync.syncing Remote digest: 67550-1519719998
2018-05-08 07:57:16,837 [DEBUG]     kitovu.sync.filecache Discovering changes for local: /home/florian/proj/kitovu/synctest/Engineering-Projekt/FS2018/Projektantraege/Projektantrag_kitovu.pdf / remote: Informatik/Fachbereich/Engineering-Projekt/EPJ/FS2018/Projektantraege/Projektantrag_kitovu.pdf by plugin smb
2018-05-08 07:57:16,837 [DEBUG]     kitovu.sync.filecache Local path does not exist!
[...]

PR #27, opened 2018-05-08
Refactor syncing.py and add PluginOperationError

The-Compiler:
See [EPJ-51](https://jira.keltec.ch/jira/browse/EPJ-51) and [EPJ-75](https://jira.keltec.ch/jira/browse/EPJ-75). Based on #26 and #25 so you might want to merge those first.

### DIFF
--- a/src/kitovu/sync/syncing.py
+++ b/src/kitovu/sync/syncing.py
@@ -62,25 +62,41 @@ def start(connection_name: str, connection_settings: ConnectionSettings) -> None
     logger.info(f'Syncing connection {connection_name}')
 
     plugin = _load_plugin(connection_settings)
-    plugin.configure(connection_settings.connection)
-    plugin.connect()
+
+    try:
+        plugin.configure(connection_settings.connection)
+        plugin.connect()
+    except utils.PluginOperationError as e:
+        logger.error(f'Error from {plugin.NAME} plugin: {e}, skipping this plugin')
 
     filecache_path: pathlib.Path = pathlib.Path(appdirs.user_data_dir('kitovu')) / 'filecache.json'
     cache: filecache.FileCache = filecache.FileCache(filecache_path)
     cache.load()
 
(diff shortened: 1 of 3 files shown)


src/kitovu/sync/syncing.py
```diff
@@ -56,70 +56,93 @@ def start_all(config_file: typing.Optional[pathlib.Path]) -> None:
...
     logger.info(f'Syncing connection {connection_name}')
 
     plugin = _load_plugin(connection_settings)
-    plugin.configure(connection_settings.connection)
-    plugin.connect()
+
+    try:
+        plugin.configure(connection_settings.connection)
+        plugin.connect()
+    except utils.PluginOperationError as ex:
```
ThunderKey, line 66: What would the use case for that be? In what case should the connection "fail" but still continue to sync the plugin files?

Shouldn't we wrap a `try`-`except UsageError` around the `start` method? Otherwise errors that should stop the plugin from continuing (eg. server not found) stop the entire sync process.

ThunderKey (changes requested):

src/kitovu/sync/syncing.py
The-Compiler: There was a missing `return` here, which I added now.

I don't think plugins should use `UsageError` at all (or only in cases where they really want to abort everything, for some reason) - also see my changes to the documentation above.

That way, you can tell plugin authors to simply use `PluginOperationError` and kitovu will do the right thing, instead of needing to tell them to use `UsageError` in some places and `PluginOperationError` in other places.

ThunderKey: Alright, we can do it that way. This way it's not possible to cancel the execution in the other methods (eg. you're in a train and loose connection, it will print "No connection to server" for each remaining file). But I think we can live with that.

ThunderKey (approved):

PR #30, opened 2018-05-08
Feature handle new local files

ThunderKey:
Tested with:
* Removing the cache file: `rm ~/.local/share/kitovu/filecache.json`
* Changing a file
* `kitovu sync`
* Everything has the state `FileState.NO_CHANGES` and does nothing (but update teh filecache)
* One Exception is the edited file which has `FileState.BOTH_CHANGED` and gets downloaded

Maybe we could make the `filecache.File` class wrapping `local_digest` and `remote_digest`. We then could move the `_compare_digests` to that class as well instead of having the `FileCache` do everything

### DIFF
--- a/src/kitovu/sync/filecache.py
+++ b/src/kitovu/sync/filecache.py
@@ -158,7 +158,10 @@ class FileCache:
         remote_digest: str = plugin.create_remote_digest(remote_full_path)
         local_digest: str = plugin.create_local_digest(local_full_path)
 
-        # Case 5 if both change to the same values
+        # If both the remote and local files are updated but the cache didn't realize it.
+        # remote = B, local = B, cache A => update the cache to B
+        # eg. Downloaded the file not via kitovu
         if remote_digest == local_digest and file.cached_digest != remote_digest:
             file.cached_digest = remote_digest
+
         return self._compare_digests(remote_digest, local_digest, file.cached_digest)
--- a/src/kitovu/sync/filecache.py
+++ b/src/kitovu/sync/filecache.py
@@ -68,10 +68,11 @@ class FileState(enum.Enum):
 @attr.s
 class File:
 
-    cached_digest: str = attr.ib()  # local digest at synctime
+    cached_digest: typing.Optional[str] = attr.ib()  # local digest at synctime
     plugin_name: str = attr.ib()
 
     def to_dict(self) -> typing.Dict[str, str]:
+        assert self.cached_digest is not None
         return {"plugin": self.plugin_name,
                 "digest": self.cached_digest}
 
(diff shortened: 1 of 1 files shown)


src/kitovu/sync/filecache.py
```diff
@@ -143,6 +143,8 @@ def discover_changes(self,
         if not local_full_path.exists():
             return FileState.NEW
 
+        if local_full_path not in self._data:
+            self._data[local_full_path] = File(cached_digest='', plugin_name=plugin.NAME)
```
The-Compiler: mypy fails with `Argument "plugin_name" to "File" has incompatible type "Optional[str]"; expected "str"` here, but an `assert plugin.NAME is not None` above (probably fine to do it even outside of the `if`) will help.

The-Compiler: Not sure if it's a good idea to use `''` for the cached digest, that makes it look like it's a bug in digest calculation when you don't know about it. What about using `None` instead?

The-Compiler: Also, please add a newline after this line, as it's unrelated to the rest of the code creating `file` (makes it easier to scan, IMHO).

src/kitovu/sync/filecache.py
```diff
@@ -152,4 +154,8 @@ def discover_changes(self,
 
         remote_digest: str = plugin.create_remote_digest(remote_full_path)
         local_digest: str = plugin.create_local_digest(local_full_path)
+
+        # Case 5 if both change to the same values
+        if remote_digest == local_digest and file.cached_digest != remote_digest:
```
The-Compiler: I'm not sure if this is the correct approach, but it's too hot for me to think properly right now :sweat_smile:
Need to think a bit more about it - wouldn't it be better to only overwrite the cached digest if we know we don't have any (i.e. it's `''` or `None`)?

@bytinbit maybe you can take a look too?

The-Compiler (changes requested):

src/kitovu/sync/filecache.py
ThunderKey: Good idea with setting the cached_digest to `None`. That way it's clear if it from the plugin or not (beacause plugins can only return strings). Additionally I've added a `assert cached_digest is not None` in the `to_dict` method to ensure it won't get saved that way (either download => local_digest or set it to the current if the `remote_digest` and `local_digest` are equal

src/kitovu/sync/filecache.py
The-Compiler: The comment should make this situation clear somehow:

image

The-Compiler:
LGTM now!

The-Compiler (approved):

The-Compiler:
Oh, something I forgot: It'd be nice to add a test for this...

ThunderKey:
Added tests in #35

PR #34, opened 2018-05-09
Add config validation to the GUI

The-Compiler:
The message is pretty ugly, but better than nothing:

image

### DIFF
--- a/src/kitovu/gui/confscreen.py
+++ b/src/kitovu/gui/confscreen.py
@@ -1,10 +1,12 @@
 import functools
 import pathlib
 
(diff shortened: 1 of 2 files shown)


ThunderKey:
Cool! Is it possible enable the user to change the size of the window? I guess this would help.

The-Compiler:
Turns out the config I used in tests before was invalid, which was breaking the tests now :laughing: - fixed!

I've looked into making the window resizable or automatically adjusting the width, but everything I could find was rather hacky. In the long run I think we should have something like a list of errors and "details" to fold out (hidden by default).

However, I think right now this is an okay-ish stop-gap solution until then.

ThunderKey (approved):

PR #35, opened 2018-05-09
added tests for when the cached digest is outdated

ThunderKey:
Tests for EPJ-79


The-Compiler (approved):
Thanks for the follow-up!

PR #36, opened 2018-05-10
Display proper URL for moodle password prompt

The-Compiler:
Quick follow-up on better password prompts - we should display the configured URL here.


ThunderKey (approved):
That's a good idea!

PR #37, opened 2018-05-10
print a better error message when the yaml file is invalid

ThunderKey:


src/kitovu/sync/settings.py
```diff
@@ -84,7 +84,10 @@ def from_yaml_stream(cls, stream: typing.IO,
             validator = utils.SchemaValidator()
 
         # FIXME handle OSError and UnicodeDecodeError
-        data = yaml.load(stream)
+        try:
+            data = yaml.load(stream)
+        except yaml.scanner.ScannerError as error:
+            raise utils.UsageError(f"Invalid Configuration:\n{str(error)}")
```
The-Compiler: Formatting things into strings automatically stringifies them, so no need to add a `str()`

The-Compiler (changes requested):
Can you please also add a quick test for this?

src/kitovu/sync/settings.py
ThunderKey: Oh yeah, right. I've removed it and added a test

The-Compiler (approved):

PR #38, opened 2018-05-10
merge global_ignore into the ignore values and actually ignore those files

ThunderKey:
EPJ-97

### DIFF
--- a/src/kitovu/sync/syncing.py
+++ b/src/kitovu/sync/syncing.py
@@ -83,7 +83,7 @@ def start(connection_name: str, connection_settings: ConnectionSettings) -> None
     plugin.disconnect()
 
 
-def _sync_subject(subject: typing.Dict[str, str],
+def _sync_subject(subject: utils.JsonSchemaType,
                   plugin: AbstractSyncPlugin,
                   cache: filecache.FileCache) -> None:
     logger.info(f'Syncing subject {subject["name"]}')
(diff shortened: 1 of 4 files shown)


src/kitovu/sync/syncing.py
```diff
@@ -91,7 +91,12 @@ def _sync_subject(subject: typing.Dict[str, str],
     remote_dir = pathlib.PurePath(subject['remote-dir'])  # /Informatik/Fachbereich/EPJ/
     local_dir = pathlib.Path(subject['local-dir'])  # /home/leonie/HSR/EPJ/
 
+    ignore = subject['ignore']
```
The-Compiler: Maybe add a type?

src/kitovu/sync/syncing.py
```diff
@@ -91,7 +91,12 @@ def _sync_subject(subject: typing.Dict[str, str],
...
+
     for remote_full_path in plugin.list_path(remote_dir):
+        if remote_full_path.parts[-1] in ignore:
```
The-Compiler: You should be able to use `.name` instead of `.parts[-1]` ([docs](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name))

The-Compiler (changes requested):

src/kitovu/sync/syncing.py
```diff
@@ -83,15 +83,20 @@ def start(connection_name: str, connection_settings: ConnectionSettings) -> None
     plugin.disconnect()
 
 
-def _sync_subject(subject: typing.Dict[str, str],
+def _sync_subject(subject: utils.JsonSchemaType,
```
The-Compiler, line 86: Strictly speaking this is wrong because we only used `JsonSchemaType` for things which are actually a JSON schema - but I think I'll unify `JsonSchemaType` and `JsonType` in `moodle.py` into one.

The-Compiler (approved):

PR #39, opened 2018-05-10
Tests for Moodle-Plugin completed

bytinbit:


src/kitovu/sync/plugin/moodle.py
```diff
@@ -51,11 +51,18 @@ def _request(self, func: str, **kwargs: str) -> typing.Any:
...
         data: JsonType = req.json()
         logger.debug(f'Got data: {data}')
-
-        assert 'exception' not in data, data
-        assert 'error' not in data, data
+        self._check_json_answer(data)
         return data
 
+    def _check_json_answer(self, data: JsonType) -> None:
+        if "errorcode" in data and data["errorcode"] == "invalidtoken":
```
The-Compiler: You can use `data.get("errorcode", None)` (`None` as fallback) here - however, since you need it twice, I'd make it a variable before the `if` anyways.

src/kitovu/sync/plugin/moodle.py
```diff
@@ -51,11 +51,18 @@ def _request(self, func: str, **kwargs: str) -> typing.Any:
...
+            raise utils.AuthenticationError(data['message'])
+        elif "errorcode" in data and data["errorcode"] == "invalidrecord":  # e.g. invalid ws_function, invalid course ID
+            # the given error message is hard to understand, writing a better to understand one instead
+            raise utils.PluginOperationError("You requested something from Moddle which it couldn't get.")
```
The-Compiler: Moddle -> Moodle

tests/assets/invalid_courseid.json
```diff
@@ -0,0 +1,20 @@
...
+Connection: Keep-Alive
+Content-Length: 164
+Content-Type: application/json
+Date: Sat, 12 May 2018 05:53:02 GMT
+Expires: Thu, 01 Jan 1970 00:00:00 GMT
+Keep-Alive: timeout=5, max=100
+Pragma: no-cache
+Server: Apache
+Strict-Transport-Security: max-age=15768000
+X-Frame-Options: sameorigin
```
The-Compiler: Those HTTP headers shouldn't be here.

tests/assets/invalid_httprequest_insteadof_https.html
```diff
@@ -0,0 +1,7 @@
+<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
```
The-Compiler: Seems like this is not used at all? If you still want to write a test using this, you'll also need to tell responses to use the correct status code (301).

tests/assets/invalid_url.html
```diff
@@ -0,0 +1,7 @@
+<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
```
The-Compiler: Seems like this is not used at all? If you still want to write a test using this, you'll also need to tell responses to use the correct status code (404).

tests/conftest.py
```diff
@@ -22,3 +22,8 @@ def temppath(tmpdir):
 @pytest.fixture
 def plugin(temppath) -> dummyplugin.DummyPlugin:
     return dummyplugin.DummyPlugin(temppath)
+
+
+@pytest.fixture
+def assets_dir() -> pathlib.Path:
```
The-Compiler, line 28: I think we should have a `moodle_assets_dir` fixture in `test_moodle.py` which adds a `moodle` subfolder to this - right now it's quite a chaos in `tests/assets`.

tests/sync/test_moodle.py
```diff
@@ -0,0 +1,298 @@
...
+@pytest.fixture
+def patch_get_wrong_wsfunction(responses):
+    body: str = """
+    {
+        "errorcode": "invalidrecord",
+        "exception": "dml_missing_record_exception",
+        "message": "Datensatz kann nicht in der Datenbanktabelle external_functions gefunden werden"
+    }
+    """
+    _patch_request(responses, 'mistake_webservice_get_site_info', body=body, wstoken="some_token")
```
The-Compiler: You need to patch the `wsfunction` which the client actually calls here.

tests/sync/test_moodle.py
```diff
@@ -0,0 +1,298 @@
...
+@pytest.fixture
+def patch_get_wrong_courseid(responses):
+    body: str = """
+        {
+        "errorcode": "invalidrecord",
+        "exception": "dml_missing_record_exception",
+        "message": "Datensatz kann nicht in der Datenbanktabelle external_functions gefunden werden"
+    }
+    """
+    _patch_request(responses, 'core_course_get_contents', body=body, courseid=424242)
```
The-Compiler: You need to patch the `courseid` which the client actually calls here.

tests/sync/test_moodle.py
```diff
@@ -0,0 +1,298 @@
...
+            '0-1487838705',
+            '0-1520427130',
+            '2119487-1490374823',
+            '44733-1396277827'
+        ]
+        for item in course_contents:
+            remote_digests.append(plugin.create_remote_digest(item))
+        assert remote_digests == check_digests
+
+    def test_list_path_with_wrong_remote_dir(self, plugin, connect_and_configure_plugin, temppath):
```
The-Compiler: This test is failing at the moment - it writes a config which isn't used in the test at all, and then we call `list_path` with a path that *does* exist?

The-Compiler (changes requested):

The-Compiler (approved):

PR #40, opened 2018-05-13
create the config file in the edit if it does not exist

ThunderKey:


src/kitovu/cli.py
```diff
@@ -99,6 +99,13 @@ def edit(config: typing.Optional[pathlib.Path] = None, editor: typing.Optional[s
 
     if config is None:
         config = settings.get_config_file_path()
+        if not config.exists():
+            if not config.parent.exists():
```
The-Compiler: Both `.mkdir()` and `.touch()` have an `exists_ok=True` parameter (it's even `True` by default for `.touch()`), so you don't need these checks with that.

The-Compiler (changes requested):
Not sure if this is needed at all - I'd expect you can easily pass a path which doesn't exist to an editor?

src/kitovu/cli.py
ThunderKey: Ah, `exist_ok=True` makes more sense.

For example vim is not able to create the file if the folder `~/.config/kitovu` doesn't exist. So we could only create the folder, but I guess there are editors which require the file to exist (I think of not-so-smart Windows editors :wink:)

lroellin:
This being open-source and me being a very interested future user, I like to chime in with @The-Compiler. This is something an editor should handle, not kitovu!

I even think it would be bad to do so. Many times I've opened vi when wanting to open a file, but tab-completion didn't find an existing file yet (ambiguous) and I wouldn't like it if a file like that would be created ;)

This being an EPJ feel *very* free to ignore my comments, but I had to chime in there, since I think it would be a design fault...

The-Compiler (approved):

The-Compiler:
Thanks! You're right, there are probably editors which don't handle this *cough* Windows *cough* :wink:
As it's only the default config file location (and not if a custom `--config` has been given), I think it's a good idea to do this.

PR #41, opened 2018-05-13
added a -h shortcut for the cli

ThunderKey:


The-Compiler:
Just pushed a commit to simplify this a bit - I don't see the need for a variable we only use once, and we use `{'foo': 'bar'}` (and not `dict(foo='bar')`) to create dicts elsewhere so we should do so here, too.

The-Compiler (approved):

PR #43, opened 2018-05-14
Improve handling of YAML errors

The-Compiler:
Fix EPJ-99 and handle all YAML errors equally


The-Compiler:
Just pushed another commit to handle some other possible errors when loading the config.

ThunderKey (approved):

PR #44, opened 2018-05-15
always load the config file in when the config edit window is opened

ThunderKey:
Before it kept the edited configuration and when you press "Abbrechen" and return to the config file it's still updated.

Hopefully I didn't break Qt conventions ;)


The-Compiler (approved):
I renamed `open_conf_screen` to `on_conf_pressed` to be consistent with `on_sync_pressed` - LGTM otherwise!

PR #46, opened 2018-05-16
More tests

The-Compiler:


.coveragerc
```diff
@@ -13,4 +13,5 @@ parallel = true
 [report]
 exclude_lines =
     pragma: no cover
-    raise NotImplementedError
\ No newline at end of file
+    raise NotImplementedError
+    raise AssertionError
```
The-Compiler, line 17: Those should never happen, so no point in covering them.

src/kitovu/gui/mainwindow.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, parent: QWidget = None) -> None:
 
         self._conf_screen.close_requested.connect(
             lambda: self.setCurrentWidget(self._start_screen))
-        self._sync_screen.close_requested.connect(
+        self._sync_screen.close_requested.connect(  # pragma: no branch
```
The-Compiler, line 28: Due to https://bitbucket.org/ned/coveragepy/issues/90/lambda-expression-confuses-branch

src/kitovu/sync/settings.py
```diff
@@ -110,25 +110,22 @@ class Settings:
     }
 
     @classmethod
-    def from_yaml_file(cls, path: typing.Optional[pathlib.Path] = None,
-                       validator: typing.Optional[utils.SchemaValidator] = None) -> 'Settings':
```
The-Compiler, line 114: I removed the ability to pass a custom validator because we didn't do that anywhere anyways (not even in tests).

tests/sync/test_settings.py
```diff
@@ -1,11 +1,38 @@
...
+    spawner.edit()
+    assert default_config.exists()
+
+
+def test_load_unreadable_file(temppath):
+    settings = temppath / 'kitovu.yaml'
+    settings.touch()
+    settings.chmod(0)
+    if os.access(settings, os.R_OK):
+        pytest.skip("Failed to make file unreadable")  # pragma: no cover
```
The-Compiler, line 31: This can happen in Docker containers or weird filesystems, so we just skip the rest of the test there.

tests/sync/test_settings.py
```diff
@@ -1,11 +1,38 @@
...
 from kitovu.sync.settings import Settings, ConnectionSettings
 
 
+@pytest.fixture
+def default_config(monkeypatch, temppath):
+    monkeypatch.setattr(settings.appdirs, 'user_config_dir', lambda _name: str(temppath))
+    return temppath / 'kitovu.yaml'
+
+
+def test_creating_default_config(default_config, monkeypatch):
```
The-Compiler, line 18: The rest of `EditorSpawner` is tested in `test_cli.py`, but this was easier to test here.

tests/sync/test_smb.py
```diff
@@ -1,17 +1,26 @@
...
-from smb.smb_structs import OperationFailure
+from smb.smb_structs import OperationFailure, ProtocolError
 
 from kitovu.sync import syncing
 from kitovu import utils
 from kitovu.sync.plugin import smb
 
 
+@attr.s
+class FakeStatResult:
```
The-Compiler, line 18: Copied from `test_moodle.py`, but I think it doesn't make much sense to move this to a helper module for this little code.

ThunderKey (approved):

PR #68, opened 2018-09-12
Get rid of Pipenv

The-Compiler:
I consider using Pipenv a failed experiment, and it made many things much harder rather than easier.

This makes tox faster, virtualenvs less confusing (as we create them by hand), and also sets up pyup.me to create pull requests for package updates in the future.

@ThunderKey I haven't tested the Docker stuff, can you please check whether that still works like expected?


ThunderKey:
I'd say we also remove the `dockerfiles`. It would be very neat if there would just be a container with which you could sync the files, but the authentication is very difficult. I couldn't get the keyring of the host to work within the container and if we save it within the container it will most likely be removed.

ThunderKey:
I think if we would give the docker container another go it would be a good idea to make a separate repo. Docker only alows to copy files in the current folder or subfolders, due to security issues. We could then add the kitovu as a submodule or even download it within the container from the pip repo.

The-Compiler:
That works for me as well, removed them. FWIW using [pyenv](https://github.com/pyenv/pyenv) (if you don't have Python 3.6) plus [virtualenv](https://virtualenv.pypa.io/en/stable/) seems to work nicely.

ThunderKey (approved):

The-Compiler:
Huh, looks like something on Windows is broken now:

> INTERNALERROR> ImportError: DLL load failed: %1 is not a valid Win32 application.
93

Will take a look some day, don't feel like debugging Windows stuff right now...

ThunderKey:
I thought the same thing. I've wanted to fix it before approving, but I couldn't find the time^^

PR #71, opened 2018-09-18
Gui import check

### LINKED ISSUE #70: PyQT5 is not a requirement
```
pyenv virtualenv 3.7.0 kitovu-3.7.0
pyenv activate kitovu-3.7.0
pip3 install kitovu
kitovu gui
```
results in
```
...
ModuleNotFoundError: No module named 'PyQt5'
```

Is just adding it to setup.py enough?

lroellin:
This fixes #70.

I'm not sure if the style of the user message is good, and if the code should be somewhere else, but I'm quite content the way it is.

I haven't written any tests of it, since I don't know how you would go about faking having/not having this module installed.

One small problem: at least in my shell (zsh) I had to escape the `[` and `]` to get it to install properly, don't know if we should warn users of this. On the other hand, it should be clear that the message comes from zsh and not the package.

I've just tested with `pip install .[gui]` (no space behind that dot), and that worked.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,7 @@ setup(
         'jsonschema',
     ],
     extras_require={
-        'gui': ['PyQt5']
+        'gui': ['PyQt5'],
     },
     entry_points={
         'console_scripts': [
--- a/setup.py
+++ b/setup.py
@@ -68,6 +68,7 @@ setup(
         'jsonschema',
     ],
     extras_require={
+        'gui': ['PyQt5']
         # eg:
         #   'rst': ['docutils>=0.11'],
         #   ':python_version=="2.6"': ['argparse'],
--- a/setup.py
+++ b/setup.py
@@ -69,9 +69,6 @@ setup(
     ],
     extras_require={
         'gui': ['PyQt5']
-        # eg:
-        #   'rst': ['docutils>=0.11'],
-        #   ':python_version=="2.6"': ['argparse'],
     },
     entry_points={
         'console_scripts': [
--- a/src/kitovu/cli.py
+++ b/src/kitovu/cli.py
@@ -44,8 +44,15 @@ def cli(loglevel: str) -> None:
 @cli.command()
 def gui() -> None:
     """Start the kitovu GUI."""
-    from kitovu.gui import app as guiapp
-    sys.exit(guiapp.run())
+    try:
+        from kitovu.gui import app as guiapp
+        sys.exit(guiapp.run())
+    except ModuleNotFoundError as e:
+        if e.name == 'PyQt5':
+            print('To run the GUI, you need to install the extra GUI dependencies')
+            print('Run "pip install kitovu[gui]"')
+        else:
+            raise
 
 
 @cli.command()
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -68,6 +68,7 @@ def read(*names, **kwargs):
         'jsonschema',
     ],
     extras_require={
+        'gui': ['PyQt5']
         # eg:
         #   'rst': ['docutils>=0.11'],
         #   ':python_version=="2.6"': ['argparse'],
```
The-Compiler: Can you remove this comment while you're at it?

setup.py
```diff
@@ -68,6 +68,7 @@ def read(*names, **kwargs):
         'jsonschema',
     ],
     extras_require={
+        'gui': ['PyQt5']
```
The-Compiler: nitpick: Add a trailing comma, so it's easier to add new entries in the future

src/kitovu/cli.py
```diff
@@ -44,8 +44,15 @@ def cli(loglevel: str) -> None:
...
     """Start the kitovu GUI."""
-    from kitovu.gui import app as guiapp
-    sys.exit(guiapp.run())
+    try:
+        from kitovu.gui import app as guiapp
+        sys.exit(guiapp.run())
+    except ModuleNotFoundError as e:
+        if e.name == 'PyQt5':
+            print('To run the GUI, you need to install the extra GUI dependencies')
+            print('Run "pip install kitovu[gui]"')
```
The-Compiler: Add `, file=sys.stderr` to both, so this goes to standard error instead of standard output.

Also, add `sys.exit(1)` so the exit status makes it clear that there was a failure.

src/kitovu/cli.py
```diff
@@ -85,8 +92,8 @@ def docs() -> None:
 @cli.command()
 @click.option('--config', type=pathlib.Path, help="The configuration file to edit")
 @click.option('--editor', type=str, help="The command of the editor to use. "
-              f"Default: $EDITOR or the first existing out of "
-              f"{settings.EditorSpawner.DEFAULT_EDITORS_STR}")
+                                         f"Default: $EDITOR or the first existing out of "
+                                         f"{settings.EditorSpawner.DEFAULT_EDITORS_STR}")
```
The-Compiler: Unrelated change? :wink:

src/kitovu/cli.py
The-Compiler: We should probably quote the `kitovu[gui]` to make things clearer - probably using double-quotes though, so it also works on Windows.

What about:

`'To do so, run: pip install "kitovu[gui]"'`

The-Compiler (changes requested):

src/kitovu/cli.py
```diff
@@ -44,8 +44,15 @@ def cli(loglevel: str) -> None:
 @cli.command()
 def gui() -> None:
     """Start the kitovu GUI."""
-    from kitovu.gui import app as guiapp
-    sys.exit(guiapp.run())
+    try:
+        from kitovu.gui import app as guiapp
+        sys.exit(guiapp.run())
+    except ModuleNotFoundError as e:
```
The-Compiler: pylint doesn't like `e` as name, looks like we use `ex` elsewhere in the codebase.

lroellin:
Should be fixed! Feedback very appreciated :)

src/kitovu/cli.py
lroellin: Pycharm Auto format ;) but I fixed it

The-Compiler:
LGTM now, but I wonder where the @codecov bot went... Anyways, imports are very hard to test properly (as patching python's import system also means pytest can't import things it needs :laughing:), so I'm fine with merging this without tests.

Thanks! :+1:

The-Compiler (approved):

PR #73, opened 2019-08-08
switch from deprecated yaml.load without a specific loader to a helper method

ThunderKey:
Issue: https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

### DIFF
--- a/src/kitovu/sync/settings.py
+++ b/src/kitovu/sync/settings.py
@@ -128,7 +128,7 @@ class Settings:
         validator = utils.SchemaValidator()
 
         try:
-            data = yaml.full_load(stream)
+            data = yaml.load(stream, Loader=yaml.FullLoader)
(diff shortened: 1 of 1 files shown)


The-Compiler:
Looks like tests fail - probably need to update PyYAML via pipenv (in `Pipfile.lock`) as well? Or maybe use `yaml.load(input, Loader=yaml.SafeLoader)` (untested) so it continues to work with older PyYAML versions too.

ThunderKey:
Good catch. I've added the older version with the `Loader` parameter.

The-Compiler:
`yaml.Loader` is the unsafe one though, and I don't think we need that? `yaml.SafeLoader` should work with both versions as well and not load any Python code.

ThunderKey:
Thanks for the clarification. I had trouble with the documentation and what the different loaders actually do, so I kept the default behaviour. I've changed that now.

PR #74, opened 2019-08-08
remove the password in the smb plugin if the authentication failed

ThunderKey:


src/kitovu/utils.py
```diff
@@ -29,6 +29,19 @@ def get_password(plugin: str, identifier: str, prompt: str) -> str:
     return password
 
 
+def delete_password(plugin: str, identifier: str) -> None:
+    """Get the password for the given URL via keyring.
```
lroellin: Probably need to update this comment too :)

ThunderKey: Whups, thanks for the hint. I've changed that.

